PR #30, opened 2020-08-05
Option to skip "name this contact" screen when send a TX to an unknown address.

svub:
Added switch to setting panel that enables to skipping the labeling screen "name this contact" when sending a TX to a previously unknown address. Why? It's a hassle if one is mostly sending to one-time addresses, e.g. to companies, collaborators, etc. in a professional setup - so such a custom setting comes handy.

### DIFF
--- a/src/components/layouts/Settings.vue
+++ b/src/components/layouts/Settings.vue
@@ -48,7 +48,11 @@
                         </p>
                     </div>
 
-                    <select id="askForLabeling" name="askForLabeling" @input="setAskForRecipientLabeling(!!$event.target.value)">
+                    <select
+                        id="askForLabeling"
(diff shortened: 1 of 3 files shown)


julianbauer:
@svub generally absolutely understandable use-case, definitely is quicker without that step. The issue is that the "Name this contact" screen is our confirmation screen for the address, to check the identicon, confirm that this is the right address. We just _also_ use it to get people to label the address, which is not required.

If we leave that step away we risk to have confusion towards the next step and people re-checking the address afterwards.

Furthermore, I definitely see a use-case in also naming one-time addresses: Having a label in the TX list to whom I sent a certain transaction helps understand my TX history better than just having addresses – even if it's just single-time payments.

PR #32, opened 2020-08-05
Adding cross-close-button to settings panel

svub:
Added a little (x) in the upper right of the settings to go back to the home screen conveniently.


sisou (approved):

sisou:
I moved the close button into the `.mobile-menu-bar` to only appear when the sidebar is hidden (tablet size and below). That way it doesn't float somewhere in the middle of the screen on desktop and makes more sense, I believe.

PR #34, opened 2020-09-07
fix & align currency flags

julianbauer:
- Updated all currency flags to match the flag-set (there were some legacy flags)
- Fixed background overlaps that were causing ugly outlines in chrome


src/assets/currencies/aud.svg
```diff
@@ -1,22 +1,50 @@
...
+<polygon class="st1" points="403,250.8 408.1,266.6 424.7,266.6 411.3,276.4 416.4,292.2 403,282.5 389.5,292.2 394.7,276.4 
+	381.2,266.6 397.8,266.6 "/>
+<polygon class="st2" points="0,152 104,152 104,256 152,256 152,152 256,152 256,104 152,104 152,0 104,0 104,104 0,104 "/>
+<g>
+	<polygon class="st2" points="167,167 256,256 256,230.8 192.1,167 	"/>
+	<polygon class="st2" points="63.9,167 0,230.8 0,256 0,256 89,167 	"/>
+</g>
+<g>
+	<polygon class="st2" points="89,89 0,0 0,25.2 63.9,89 	"/>
+	<polygon class="st2" points="192.1,89 256,25.2 256,0 256,0 167,89 	"/>
```
sisou, line 48: I think these coordinates have unnecessary many decimals, plus there's excess whitespace in the last 6 lines.

src/assets/currencies/brl.svg
```diff
@@ -1,9 +1,32 @@
...
+		c12.4-28.8,41-48.9,74.3-48.9c44.6,0,80.8,36.2,80.8,80.8C336.8,260.2,336.5,264.3,335.9,268.3z"/>
+</g>
+<path class="st3" d="M256,114.5L64,256l192,141.5L448,256L256,114.5z M256,336.8c-44.6,0-80.8-36.2-80.8-80.8s36.2-80.8,80.8-80.8
+	s80.8,36.2,80.8,80.8S300.6,336.8,256,336.8z"/>
+<g>
+	<path class="st2" d="M215.6,250.9c-14.1,0-27.6,2.1-40.4,6.1c0.6,44.2,36.5,79.8,80.8,79.8c27.4,0,51.6-13.6,66.2-34.5
+		C297.2,271.1,258.7,250.9,215.6,250.9z"/>
+	<path class="st2" d="M335.3,271.5c1-5,1.5-10.2,1.5-15.5c0-44.6-36.2-80.8-80.8-80.8c-33.3,0-61.9,20.2-74.3,48.9
+		c10.9-2.3,22.3-3.5,33.9-3.5C262.5,220.6,305,240.1,335.3,271.5z"/>
+</g>
```
sisou, line 31: This is unnecessary big. A simple pass through SVGO (https://jakearchibald.github.io/svgomg/) reduces this to half the size.

sisou (changes requested):
Thanks for the update. However, most of these "updated" flags are unnecessarily big and need a pass through SVGO (https://jakearchibald.github.io/svgomg/) to get optimized. They are displayed at such a small scale, that most often the highest optimization can be applied and users won't see the difference, but they feel it loading faster and we don't ship unnecessary bytes :pray:

PR #42, opened 2021-02-20
fix: overflowing tooltips on network map

### LINKED ISSUE #15: Network map tooltips overflowing
When network map tooltips become long because of many nodes at that hexagon, they can overflow the window because they are oriented to the wrong direction:

User-submitted bug image

mariofriz:
This PR fixes #15 

## Description
Break down the map in 4 squares and depending on the position of the hexagon on the map, set the tooltip position instead of relying on auto-positioning.

image

## Changes
- calculate `preferredPosition` for each `NodeHexagon` on the map

### DIFF
--- a/src/components/NetworkMap.vue
+++ b/src/components/NetworkMap.vue
@@ -7,6 +7,7 @@
                 class="node"
                 theme="inverse"
                 :container="$container ? { $el: $container } : null"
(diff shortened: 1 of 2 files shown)


sisou (approved):
This looks great!
We'll try it out and merge it for the next release in a few days! :+1:

PR #43, opened 2021-02-20
fix: fetch account history on re-login

### LINKED ISSUE #27: Account not synced on re-login
Steps to reproduce:
- Be logged-in in account A and account B
- Logout from B
- Login into B again
-> The balance and transaction history for B is not being loaded (this is better visible if the accounts share no common transaction history).

mariofriz:
This PR fixes #27 

## Description

The account history is only fetched once per address. This causes a problem when login out and login in again because history is removed during logout but not fetched again during login. 

**Steps to reproduce**
- Log into 2 accounts that both have account history
- Log out of one account
- Log in again with that account
- Account history is not fetched for that account (only transactions that are in both accounts are displayed)

## Changes
- in `addressStore.addressInfos` watcher, track addresses that have been removed (logout)
- delete the `removedAddresses` from the `fetchedAddresses` so they are fetched again once added (login)
- fixed typo "heigth" -> "height"

### DIFF
--- a/src/network.ts
+++ b/src/network.ts
@@ -104,6 +104,12 @@ export async function launchNetwork() {
         client.getTransactionsByAddress(address, /* lastConfirmedHeight - 10 */ 0, knownTxDetails)
             .then((txDetails) => {
                 transactionsStore.addTransactions(txDetails);
+                // update address balance because balanceListeners are not triggered
+                const { state: { transactions } } = transactionsStore;
+                const balance = Object.values(transactions)
+                    .filter((tx) => tx.sender === address || tx.recipient === address)
+                    .reduce((sum, tx) => tx.recipient === address ? sum + tx.value : sum - tx.value, 0);
+                addressStore.patchAddress(address, { balance });
             })
             .catch(() => fetchedAddresses.delete(address))
             .then(() => network$.fetchingTxHistory--);
--- a/src/network.ts
+++ b/src/network.ts
@@ -60,7 +60,7 @@ export async function launchNetwork() {
     // Also remove logged out addresses from fetched (so that they get fetched on next login)
     watch(addressStore.addressInfos, () => {
         const newAddresses: string[] = [];
-        const removedAddresses: Set<string> = new Set<string>(subscribedAddresses);
+        const removedAddresses = new Set(subscribedAddresses);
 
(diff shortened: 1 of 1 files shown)


src/network.ts
```diff
@@ -53,23 +53,36 @@ export async function launchNetwork() {
...
-        console.debug('Subscribing addresses', newAddresses);
-        client.subscribe(newAddresses);
+        if (newAddresses.length > 0) {
+            console.debug('Subscribing addresses', newAddresses);
+            client.subscribe(newAddresses);
+        }
+
+        removedAddresses.forEach((removedAddress) => {
+            fetchedAddresses.delete(removedAddress);
+        });
```
sisou: I suggest to simply handle the `removedAddresses` before the `return` condition, to not unnecessarily change code that doesn't need changing.

```ts
        removedAddresses.forEach((removedAddress) => {
            fetchedAddresses.delete(removedAddress);
        });

        if (!newAddresses.length) return;

        console.debug('Subscribing addresses', newAddresses);
        client.subscribe(newAddresses);
```

src/network.ts
```diff
@@ -53,23 +53,36 @@ export async function launchNetwork() {
...
 
-    // Subscribe to new addresses (for balance updates and transactions)
     const subscribedAddresses = new Set<string>();
+    const fetchedAddresses = new Set<string>();
+
+    // Subscribe to new addresses (for balance updates and transactions)
+    // Also remove logged out addresses from fetched (so that they get fetched on next login)
     watch(addressStore.addressInfos, () => {
         const newAddresses: string[] = [];
+        const removedAddresses: Set<string> = new Set<string>(subscribedAddresses);
```
sisou: Both type annotations are unnecessary here, as TS can infer that creating a new set from a `Set<string>` creates a `Set<string>`.

sisou (changes requested):

src/network.ts
mariofriz: Haha, I've been doing too much Java lately and unconsciously putting types everwhere!

mariofriz:
Thanks for the feedbacks @sisou, will fix this tonight

mariofriz:
@sisou Actually noticed the account and address balances are not updating. I tried to fix it but seems hacky.
Problem is that the `balancesListener` are not getting triggered so address balance remains `null`.

PR #44, opened 2021-02-20
fix: show error when sender is same as recipient

### LINKED ISSUE #19: Transaction with same recipient as sender only fails after signing
A transaction's sender and recipient cannot be the same address. The SendModal however does not detect this case and the transaction only fails to send when sending to the network, after it was signed.

mariofriz:
This PR fixes #19 

## Description

When sender and recipient are the same the transaction only fails during transaction step. It should fail early and inform user.

**Steps to reproduce**
- Copy address of current active address
- Open "Send" modal
- Paste address
- No validation happens and amount input is shown

## Changes
- Add recipient validation (sender address different from recipient address)
- Fix small indentation error

## Todo
- [x] Validate error message
- [ ] Translate error message

### DIFF
--- a/src/components/modals/SendModal.vue
+++ b/src/components/modals/SendModal.vue
@@ -83,15 +83,25 @@
                         <label>{{ activeAddressInfo.label }}</label>
                     </button>
                     <div class="separator-wrapper">
-                            <div class="separator"></div>
-                        </div>
+                        <div class="separator"></div>
+                    </div>
                     <IdenticonButton
                         :address="recipientWithLabel.address"
                         :label="recipientWithLabel.label"
                         @click="recipientDetailsOpened = true"/>
                 </section>
 
-                <section class="amount-section" :class="{'insufficient-balance': maxSendableAmount < amount}">
+                <section v-if="!isValidRecipient" class="invalid-recipient-section">
+                    <span class="nq-notice warning" key="invalid-recipient">
+                        {{ $t('Invalid recipient. The sender and recipient must be different.') }}
+                    </span>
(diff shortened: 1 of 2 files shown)


src/components/modals/SendModal.vue
```diff
@@ -141,6 +151,7 @@
                     <LabelInput
                         v-model="message"
                         :placeholder="$t('Add a public message...')"
+                        :disabled="!isValidRecipient"
                         :maxBytes="64"
```
sisou: Instead of disabling the `message` LabelInput, the whole section should be hidden, just like the `amount-section`.

sisou (changes requested):

sisou:
@Tammoovernice 
This is the wording suggested for this change. Do you agree or would you suggest differently?
image

Tammoovernice:
> @Tammoovernice
> This is the wording suggested for this change. Do you agree or would you suggest differently?
> image

Let's lose the 'Invalid recipient'. It's tech talk + we can't be sure that the user has the correct recipient but the wrong sender.
I would suggest 'The sender and recipient cannot be identical.' to avoid the more strict 'must'.

mariofriz:
Can someone please explain the i18n process to me? 

Updated the text and remove message input:
Screenshot from 2021-02-25 21 38 07

sisou (approved):
Regarding i18n: You already extracted the en.po file. After it's merged, we'll push it to Transifex so it can get translated by our translators. Then later, when translated, it will be pulled down in the next deployment.

PR #52, opened 2021-05-10
Add SRI to all scripts

paberr:
Adds SRI tags to the scripts included in the wallet.


mraveux (approved):
Thank you for the PR @paberr ! That's actually a nice thing to have, and pretty simple to setup. Thanks!

PR #61, opened 2021-07-02
Mitigate iOS 14.6 IndexedDB issue

curdbecker:
Using IndexedDB facilities in Safari always fails in iOS 14.6 on their first
use in a new browser context due to an underlying module lazy loading issue.

Hence, the mitigation waits for the IndexedDB initialisation to complete within
a timeout. If the initialisation isn't completed by then, it is retried until it succeeds.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -5,9 +5,6 @@ import VueVirtualScroller from 'vue-virtual-scroller';
 import { setAssetPublicPath as setVueComponentsAssetPath } from '@nimiq/vue-components';
 import { init as initFastspotApi } from '@nimiq/fastspot-api';
 
-// TODO: remove this as soon as the Safari IndexedDB bug is fixed
-import idbReady from 'safari-14-idb-fix';
-
 import Config from 'config';
 import App from './App.vue';
 import { serviceWorkerHasUpdate } from './registerServiceWorker';
@@ -38,8 +35,6 @@ Vue.use(VueCompositionApi);
 Vue.use(VueVirtualScroller);
 
 async function start() {
-    await idbReady(); // TODO: remove this as soon as the Safari IndexedDB bug is fixed
-
     await initStorage(); // Must be awaited before starting Vue
     await initHubApi(); // Must be called after VueCompositionApi has been enabled
     syncFromHub(); // Can run parallel to Vue initialization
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -34,8 +34,33 @@ Vue.config.productionTip = false;
...
 async function start() {
-    await initStorage(); // Must be awaited before starting Vue
+    // LocalStorage must be initalised before starting Vue
+    // FIXME: Remove this as soon the Apple fixes the Safari bug
+    try {
+        // Due to a bug in iOS 14.6 IndexedDB initalisation *always* fails
+        // on the first attempt due to some lazy module loading issue in Safari...
+        // Therefore, we've to try it twice :-/
+        // see also: https://developer.apple.com/forums/thread/681201
+        await timeout(initStorage, 1000, timeoutError);
```
sisou: I believe you need to call the function to get a promise, otherwise it always runs into the timeout.
```suggestion
        await timeout(initStorage(), 1000, timeoutError);
```

sisou (approved):
It's not nice, but necessary. Maybe it can be extracted into it's own file to not litter the `main.ts`, as I think we need to keep this around for a loooong time.

sisou:
Actually, thinking about it: since this only concerns Indexeddb, and the storage engine also supports localstorage, maybe this fix should rather be placed in `storage.ts` directly around the Indexeddb initialization call.

sisou:
There's also this https://github.com/jakearchibald/safari-14-idb-fix

src/main.ts
curdbecker: Hm, then I was unfortunately right the first time, the patch doesn't work like this. The execution then never reaches the next line (as far as I can tell). Then, it just works because of the 1s delay... 🤔

curdbecker:
> Actually, thinking about it: since this only concerns Indexeddb, and the storage engine also supports localstorage, maybe this fix should rather be placed in `storage.ts` directly around the Indexeddb initialization call.

Well, it might be just me and my limited experience, but I couldn't really find the initialisation call, so I just took "the whole thing" 😂

curdbecker:
> There's also this https://github.com/jakearchibald/safari-14-idb-fix

Yep, you're right. That's much cleaner - and it also avoids to reliably determine the point where IndexedDB is first used in the storage engine... 😅

PR #70, opened 2022-01-07
QR scanner is using full width in small screens

onmax:
Before:
before change

After:
after change


danimoh (approved):

sisou (approved):
Approved pending rebase on merged #71

sisou (approved):

PR #71, opened 2022-01-07
Update SASS compiler

sisou:
The old `node-sass` is deprecated and now produces build errors in GitHub Actions, as seen in #70.

This updates the compiler to the latest supported `sass` package and updates the relevant syntax that was throwing errors and warning post-update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "eslint-plugin-import": "^2.20.1",
     "eslint-plugin-vue": "^6.2.2",
     "node-fetch": "^2.6.1",
-    "node-sass": "^4.13.1",
+    "sass": "^1.46.0",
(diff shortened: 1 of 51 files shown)


danimoh:
Should we use `>>>` instead of `/deep/`?
I think that's what we're using everywhere else, or at least what I'm using :D

mraveux:
I never knew if we prefered to use one over the other, but i tried to use the same concistently.  I also prefer the >>> version if possible but i think the most recent one is :deep(.selector) now, isn't it @danimoh ?

sisou:
Vue 2 with SASS has to use ::v-deep
https://stackoverflow.com/a/55368933

danimoh (approved):

PR #72, opened 2022-01-07
Address selector on mobile

onmax:
## Current state

If user is using a mobile device and the account column is active, the buttons "receive" and "send" will use the last selected address as a recipient or as a sender. This is not very intuitive.

[See video](https://www.loom.com/share/d5dec9857bc54d538aeb7811a47df6d1)

## Changes in this PR

With this PR, if the user is using a mobile device and account column is active, when clicking "receive" and "send", user can select a NIM address or BTC

[See video](https://www.loom.com/share/a3d0fcc419fc429192979211c8412bcf)

## Changes in code

1. Added new composable `useActiveMobileColumn`
2. New way to close modal to support user stories that contain multiple modal one after the other
3. Added `AddressSelectorModal`
4. Refactor `/send` and `/receive` paths
5. Added back arrow to some screens if the user is coming from the address selector and updated some screens repositioning some icons

## Future work

- User in 'Choose a Sender' should not be able to select addresses with 0 assets
- In `BTCSendModal`, back arrow and info icon are shown very close to each other, it might be subject of change in the future

### DIFF
--- a/src/components/MobileActionBar.vue
+++ b/src/components/MobileActionBar.vue
@@ -18,6 +18,8 @@
 <script lang="ts">
 import { defineComponent, computed } from '@vue/composition-api';
 import { ArrowRightSmallIcon, ScanQrCodeIcon } from '@nimiq/vue-components';
+import { useWindowSize } from '@/composables/useWindowSize';
+import { ColumnType, useActiveMobileColumn } from '@/composables/useActiveMobileColumn';
 import { useAddressStore } from '../stores/Address';
 import { useAccountStore } from '../stores/Account';
 import { CryptoCurrency } from '../lib/Constants';
@@ -28,6 +30,9 @@ export default defineComponent({
         const { activeAddressInfo } = useAddressStore();
         const { activeCurrency } = useAccountStore();
         const { accountBalance } = useBtcAddressStore();
+        const { width } = useWindowSize();
+
+        const { activeMobileColumn } = useActiveMobileColumn();
 
         function nimOrBtc<T>(nim: T, btc: T): T {
             switch (activeCurrency.value) {
@@ -38,7 +43,11 @@ export default defineComponent({
         }
 
         function receive() {
-            context.root.$router.push(nimOrBtc<string>('/receive', '/btc-receive'));
+            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+                context.root.$router.push('/receive');
+            } else {
+                context.root.$router.push(nimOrBtc<string>('/receive/nim', '/receive/btc'));
+            }
         }
 
         function send() {
--- a/src/components/MobileActionBar.vue
+++ b/src/components/MobileActionBar.vue
@@ -18,7 +18,7 @@
 <script lang="ts">
 import { defineComponent, computed } from '@vue/composition-api';
 import { ArrowRightSmallIcon, ScanQrCodeIcon } from '@nimiq/vue-components';
-import { useAddressStore } from '../stores/Address';
+import { AddressType, useAddressStore } from '../stores/Address';
 import { useAccountStore } from '../stores/Account';
 import { CryptoCurrency } from '../lib/Constants';
 import { useBtcAddressStore } from '../stores/BtcAddress';
@@ -27,8 +27,8 @@ import { ColumnType, useActiveMobileColumn } from '../composables/useActiveMobil
 
 export default defineComponent({
     setup(props, context) {
-        const { activeAddressInfo } = useAddressStore();
-        const { activeCurrency } = useAccountStore();
+        const { activeAddressInfo, addressInfos } = useAddressStore();
+        const { activeCurrency, activeAccountInfo } = useAccountStore();
         const { accountBalance } = useBtcAddressStore();
         const { width } = useWindowSize();
         const { activeMobileColumn } = useActiveMobileColumn();
@@ -41,21 +41,35 @@ export default defineComponent({
             }
         }
 
+        const hasBitcoin = computed(() => Boolean(activeAccountInfo.value
+                && activeAccountInfo.value.btcAddresses && activeAccountInfo.value.btcAddresses.external.length));
+
+        const hasMultipleReceivableAddresses = computed(() => (
+            addressInfos.value.filter(({ type }) => type === AddressType.BASIC).length > 1));
+
         function receive() {
-            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+            if (width.value <= 700 /* Full mobile breakpoint */
+                && activeMobileColumn.value !== ColumnType.ADDRESS
+                && (hasMultipleReceivableAddresses.value || hasBitcoin.value)
+            ) {
                 // redirect to the address selector
                 context.root.$router.push('/receive');
             } else {
-                context.root.$router.push(nimOrBtc<string>('/receive/nim', '/receive/btc'));
+                context.root.$router.push(nimOrBtc('/receive/nim', '/receive/btc'));
             }
         }
 
+        const hasMultipleSendableAddress = computed(() => activeAccountInfo.value!.addresses.length > 1);
+
         function send() {
-            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+            if (width.value <= 700 /* Full mobile breakpoint */
+                && activeMobileColumn.value !== ColumnType.ADDRESS
+                && (hasMultipleSendableAddress.value || hasBitcoin.value)
+            ) {
                 // redirect to the address selector
                 context.root.$router.push('/send');
             } else {
-                context.root.$router.push(nimOrBtc<string>('/send/nim', '/send/btc'));
+                context.root.$router.push(nimOrBtc('/send/nim', '/send/btc'));
             }
         }
 
--- a/src/components/MobileActionBar.vue
+++ b/src/components/MobileActionBar.vue
@@ -16,14 +16,14 @@
 </template>
 
 <script lang="ts">
-import { ColumnType, useActiveMobileColumn } from '@/composables/useActiveMobileColumn';
-import { useWindowSize } from '@/composables/useWindowSize';
+import { defineComponent, computed } from '@vue/composition-api';
 import { ArrowRightSmallIcon, ScanQrCodeIcon } from '@nimiq/vue-components';
-import { computed, defineComponent } from '@vue/composition-api';
-import { CryptoCurrency } from '../lib/Constants';
-import { useAccountStore } from '../stores/Account';
 import { useAddressStore } from '../stores/Address';
+import { useAccountStore } from '../stores/Account';
+import { CryptoCurrency } from '../lib/Constants';
 import { useBtcAddressStore } from '../stores/BtcAddress';
+import { useWindowSize } from '../composables/useWindowSize';
+import { ColumnType, useActiveMobileColumn } from '../composables/useActiveMobileColumn';
 
 export default defineComponent({
     setup(props, context) {
--- a/src/components/MobileActionBar.vue
+++ b/src/components/MobileActionBar.vue
@@ -16,13 +16,13 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, computed } from '@vue/composition-api';
-import { ArrowRightSmallIcon, ScanQrCodeIcon } from '@nimiq/vue-components';
-import { useWindowSize } from '@/composables/useWindowSize';
 import { ColumnType, useActiveMobileColumn } from '@/composables/useActiveMobileColumn';
-import { useAddressStore } from '../stores/Address';
-import { useAccountStore } from '../stores/Account';
+import { useWindowSize } from '@/composables/useWindowSize';
+import { ArrowRightSmallIcon, ScanQrCodeIcon } from '@nimiq/vue-components';
+import { computed, defineComponent } from '@vue/composition-api';
 import { CryptoCurrency } from '../lib/Constants';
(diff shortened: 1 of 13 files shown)


src/components/MobileActionBar.vue
```diff
@@ -38,14 +42,24 @@ export default defineComponent({
         }
 
         function receive() {
-            context.root.$router.push(nimOrBtc<string>('/receive', '/btc-receive'));
+            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+                // redirect to the address selector
+                context.root.$router.push('/receive');
+            } else {
+                context.root.$router.push(nimOrBtc<string>('/receive/nim', '/receive/btc'));
```
sisou, line 49: The `<string>` specification of the generic type of the `nimOrBtc` function is actually not necessary, Typescript figures it out itself.

sisou: The `<string>` specification of the generic type of the `nimOrBtc` function is actually not necessary, Typescript figures it out itself.

src/components/modals/AddressSelectorModal.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+    <Modal>
+        <PageHeader>
+            <template v-if="path === '/send'">
+                {{ $t("Choose a Sender")}}
+            </template>
+            <template v-else-if="path === '/receive'">
+                {{ $t("Choose a Receiver")}}
```
sisou: Maybe "Choose a Recipient"? We can ask Elion also, he usually has an opinion on this :D

src/components/modals/AddressSelectorModal.vue
```diff
@@ -0,0 +1,65 @@
...
+            </template>
+            <template v-else>
+                {{ $t("Choose an Address")}}
+            </template>
+        </PageHeader>
+        <PageBody>
+            <AddressList
+                embedded
+                showBitcoin
```
sisou: Legacy accounts do not have Bitcoin, and iOS or Ledger accounts that have not activated it also don't have a BTC account. So this needs to be conditional, and only shown when BTC is available/enabled on this account.

src/components/MobileActionBar.vue
```diff
@@ -38,14 +42,24 @@ export default defineComponent({
         }
 
         function receive() {
-            context.root.$router.push(nimOrBtc<string>('/receive', '/btc-receive'));
+            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+                // redirect to the address selector
+                context.root.$router.push('/receive');
+            } else {
```
sisou: Accounts that only have one eligible address (read: only one regular address, contracts don't count) and don't have BTC available/enabled, we should skip the selection and go directly to /receive/nim.

src/components/MobileActionBar.vue
```diff
@@ -38,14 +42,24 @@ export default defineComponent({
...
+                context.root.$router.push(nimOrBtc<string>('/receive/nim', '/receive/btc'));
+            }
         }
 
         function send() {
-            context.root.$router.push(nimOrBtc<string>('/send', '/btc-send'));
+            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+                // redirect to the address selector
+                context.root.$router.push('/send');
+            } else {
```
sisou: If only one address exists on the account, and BTC is not available/enabled, skip the selection and go directly to /send/nim.

src/components/modals/BtcReceiveModal.vue
```diff
@@ -317,6 +324,14 @@ export default defineComponent({
             window.removeEventListener('resize', updateAddressFontSizeScaleFactor);
         });
 
+        // User only can go back to the address selection if is mobile and the column shown is the account
```
sisou:
```suggestion
        // User can only go back to the address selection if is mobile and the column shown is the account
```

src/components/modals/BtcReceiveModal.vue
```diff
@@ -317,6 +324,14 @@ export default defineComponent({
...
+        const canUserGoBack = ref(width.value <= 700 && activeMobileColumn.value !== ColumnType.ADDRESS);
```
sisou: This way of doing it works for now, but may not hold up in the future when we might have new ways to reach this modal. 

This way right now is _implicit_, where the receive modal itself determines if it should show the back button or not, depending on where it's shown. It does not consider where the user came from.

I think a better way would be to make this choice _explicit_ to the function call that opens this route. This modal can take a prop that determines if the back-arrow should be shown, the prop can be passed as a `param` in the router.push call.

src/components/modals/BtcReceiveModal.vue
```diff
@@ -362,12 +379,77 @@ export default defineComponent({
...
+        display: flex;
+        margin-top: 1.5rem;
+        justify-content: center;
+        align-items: center;
         font-size: var(--body-size);
         line-height: 1.4;
         font-weight: 600;
-        opacity: 0.6;
-        margin-top: 2rem;
+        text-align: initial;
```
sisou, line 381: Why is `text-align` needed? Removing it makes no visible difference for me.

src/components/modals/BtcSendModal.vue
```diff
@@ -494,16 +498,7 @@ export default defineComponent({
...
-                    if (window.history.state.cameFromSend) {
-                        // This is required when going to the QR scanner from within /btc-send, as a sucessful
-                        // scan _replaces_ the /scan route with the result, meaning the original /btc-send is
-                        // the previous history entry.
-                        context.root.$router.go(-2);
-                    } else {
-                        context.root.$router.back();
-                    }
-                }, SUCCESS_REDIRECT_DELAY);
+                sucessCloseTimeout = window.setTimeout(() => $modal.value!.forceClose(), SUCCESS_REDIRECT_DELAY);
```
sisou: Typo: `sucess` => `success`

src/components/modals/Modal.vue
```diff
@@ -150,8 +176,15 @@ export default defineComponent({
             close();
         }
 
+        onMounted(() => {
+            setTimeout(() => {
+                enableModalTransition();
+            }, 100);
```
sisou, line 182: Why is the timeout required here?

src/composables/useActiveMobileColumn.ts
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        next();
+    });
+}, 0);
+
+export function useActiveMobileColumn() {
+    if (!activeMobileColumn) {
+        activeMobileColumn = ref(ColumnType.ACCOUNT);
+    }
```
sisou, line 51: Since the variable is initialized here when needed and also in the beforeResolve function above, wouldn't it make sense to directly initialize it to its Ref-value, instead of to null first?

src/i18n/en.po
```diff
@@ -151,15 +151,15 @@ msgstr ""
 msgid "Add a label to quickly find the transaction in your history."
 msgstr ""
 
-#: src/components/modals/SendModal.vue:174
+#: src/components/modals/SendModal.vue:176
```
sisou: We usually try to exclude the language files from PRs, as they only create merge conflicts. They are rebuild when the app is deployed.

sisou (changes requested):
Well done :+1:

One issue in regards to how the address-selection-modal handles legacy accounts and accounts without BTC, and one optional issue where I think the logic to show the back arrow can be handled better explicitly.

src/components/modals/BtcReceiveModal.vue
danimoh: But the param would be visible in the url, which is not so nice.
It could also be stored in the history state when switching the modal.

src/components/modals/BtcReceiveModal.vue
danimoh: To align the text to the left when it breaks into two lines.

src/components/modals/Modal.vue
danimoh: At the time the modal instance gets created, the transition still needs to be disabled to avoid that showing this modal is animated.

src/composables/useActiveMobileColumn.ts
danimoh: The ref function can apparently not be called at that time yet.
cf https://github.com/nimiq/wallet/blob/master/src/composables/useWindowSize.ts#L5-L9

src/i18n/en.po
danimoh: :see_no_evil:

src/components/modals/BtcReceiveModal.vue
sisou: This is false. There is a `param` option for router methods, that pass as props, not as `query` params.
https://router.vuejs.org/guide/essentials/passing-props.html
https://router.vuejs.org/api/#route-object-properties

src/composables/useActiveMobileColumn.ts
sisou: Right. Can this comment be copied to here, too, so this is documented?

src/components/modals/BtcReceiveModal.vue
onmax: I think there are only two ways of passing variables between pages in Vue and both of them use the URI as a channel :
1. `URL params`: As an example `/user/onmax` where `onmax` could be the variable, and then in the router you indicate Vue how you want to map the value
2. `Query param`: As an example `/user?username=onmax` where username is the key that the router will use to map the value

Therefore, what you are proposing is modifying the URI to something like `/receive/nim/canUserGoBack`, and @danimoh and me discarded that option. If I am not wrong, the only way to pass data between routes in Vue is using the URI using `URL params` or `query params`

onmax:
@sisou all the change requests and suggestions have been made 🐒

src/components/modals/BtcReceiveModal.vue
sisou: We found that using router params still works when used with specifying the route by its `name`, and if no placeholder is set in the route's path, the param does not show up there, but is still available on `$router.params`.

sisou (approved):

src/components/MobileActionBar.vue
```diff
@@ -41,21 +41,36 @@ export default defineComponent({
             }
         }
 
+        const hasBitcoin = computed(() => Boolean(activeAccountInfo.value
+                && activeAccountInfo.value.btcAddresses && activeAccountInfo.value.btcAddresses.external.length));
```
danimoh, line 45: Should be updated after https://github.com/nimiq/wallet/pull/74 is merged.

src/components/MobileActionBar.vue
```diff
@@ -41,21 +41,36 @@ export default defineComponent({
...
+
+        const hasMultipleReceivableAddresses = computed(() => (
+            addressInfos.value.filter(({ type }) => type === AddressType.BASIC).length > 1));
+
         function receive() {
-            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+            if (width.value <= 700 /* Full mobile breakpoint */
```
danimoh, line 51: Should be updated after https://github.com/nimiq/wallet/pull/74 is merged.

src/components/MobileActionBar.vue
```diff
@@ -41,21 +41,36 @@ export default defineComponent({
...
+                context.root.$router.push(nimOrBtc('/receive/nim', '/receive/btc'));
             }
         }
 
+        const hasMultipleSendableAddresses = computed(() =>
+            activeAccountInfo.value && activeAccountInfo.value.addresses.length > 1);
+
         function send() {
-            if (width.value <= 700 /* Full mobile breakpoint */ && activeMobileColumn.value !== ColumnType.ADDRESS) {
+            if (width.value <= 700 /* Full mobile breakpoint */
```
danimoh, line 66: Should be updated after https://github.com/nimiq/wallet/pull/74 is merged.

src/components/modals/AddressSelectorModal.vue
```diff
@@ -32,17 +32,28 @@ export default defineComponent({
...
+                params: {
+                    canUserGoBack: 'true',
+                },
+            });
         }
 
+        const { activeAccountInfo } = useAccountStore();
+        const showBitcoin = computed(() => Boolean(activeAccountInfo.value
+            && activeAccountInfo.value.btcAddresses
+            && activeAccountInfo.value.btcAddresses.external.length));
```
danimoh, line 54: Should be updated after https://github.com/nimiq/wallet/pull/74 is merged.

src/components/modals/AddressSelectorModal.vue
```diff
@@ -32,17 +32,28 @@ export default defineComponent({
...
+        const { name } = context.root.$router.currentRoute;
 
         function addressSelected() {
             disableNextModalTransition();
 
-            context.root.$router.push(`${path}/${activeCurrency.value}`);
+            context.root.$router.push({
+                name: `${name}-${activeCurrency.value}`,
+                params: {
+                    canUserGoBack: 'true',
```
danimoh, line 46: Does `true` have to be a string?

Edit: I guess [this comment](https://github.com/nimiq/wallet/pull/72/commits/02906e011bb07416d0fbdff3f63ed0aa5b8731d8#diff-a90bbf62b9d812fd389c58b3a09c10f35a5eb609278359146245cecbb7abf06cL147) should be put here, if it still applies.
(Kind of confusing that `'false'` will lead to the same behavior).

danimoh (approved):

src/components/modals/AddressSelectorModal.vue
onmax: You are right, adding a `'false'` as a string will have the same behaviour, although I've added a comment 😸

onmax:
Just realized that there are some useless props. I remove them in the last commit 3ed4a25

PR #73, opened 2022-01-08
Add an arrow to identicon/bank selection buttons, to indicate interactivity

sisou:
This adds a little triangle to identicon-stack and bank icon buttons in NIM SendModal and the OASIS buy and sell modals, to indicate that those are "dropdowns" that can be clicked:

image
image
image

### DIFF
--- a/src/components/BankIconButton.vue
+++ b/src/components/BankIconButton.vue
@@ -1,7 +1,7 @@
 <template>
     <button class="reset bank-icon-button flex-column" v-on="$listeners">
         <BankIcon/>
-        <TriangleDownIcon class="triangle"/>
+        <TriangleDownIcon/>
         <label>{{ bankName || '' }}</label>
     </button>
 </template>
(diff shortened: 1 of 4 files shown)


src/components/icons/TriangleDownIcon.vue
```diff
@@ -0,0 +1,5 @@
+<template>
```
danimoh:
```suggestion
<template functional>
```

src/components/icons/TriangleDownIcon.vue
```diff
@@ -0,0 +1,5 @@
...
+    <svg width="10" height="8" viewBox="0 0 10 8" fill="currentColor" xmlns="http://www.w3.org/2000/svg" class="nq-icon">
```
danimoh:
```suggestion
    <svg width="10" height="8" viewBox="0 0 10 8" fill="currentColor" xmlns="http://www.w3.org/2000/svg" class="nq-icon triangle-down-icon">
```

src/components/BankIconButton.vue
```diff
@@ -1,26 +1,30 @@
 <template>
     <button class="reset bank-icon-button flex-column" v-on="$listeners">
         <BankIcon/>
+        <TriangleDownIcon class="triangle"/>
```
danimoh: I would prefer adding the class `triangle-down-icon` to the `TriangleDownIcon` component and then using that class.

src/components/BankIconButton.vue
```diff
@@ -32,9 +36,17 @@ export default defineComponent({
...
-        margin-top: -.5rem;
+    }
+
+    .triangle {
+        position: absolute;
+        right: 4rem;
+        top: 8rem;
+        width: 1.25rem;
+        height: 1rem;
+        opacity: 0.25;
```
danimoh: Would be nice to avoid the code duplication.
Maybe by moving the style `button .triangle-down-icon` to `TriangleDownIcon`?

danimoh: Add a hover effect?

src/components/IdenticonStack.vue
```diff
@@ -130,6 +135,15 @@ export default defineComponent({
         }
     }
 
+    .triangle {
+        position: absolute;
+        right: 2.5rem;
+        top: 8rem;
```
danimoh, line 142: The position does not look so great if there are only one or two addresses in the identicon selector.

danimoh (changes requested):

PR #76, opened 2022-01-19
Onboarding & Network Tour

onmax:
# Wallet Tours

This PR brings tours to the wallet.

For now, onboarding and network tour has been made, but the logic of the tour has been made in a way that it is easy to add new tour, even copy and paste it in other vue projects.

[See it in action in PC](https://www.loom.com/share/44d1e27d677743a49e4e3ba7b237c68c)

## Things that need to be done

- Review texts and wording
- Finish new `WelcomeModal` design (currently blocked by overnice)
- Check onboarding tour steps as there are some cases that might change (currently blocked by overnice)

## Features

- Mobile responsive 📱
- Key navigation ⌨️ 
- On resize events 🏃 
- Easy to create new tours 🏆 
- Each step is very customizable 🪄 
- Steps are reactive: The step will show different state/content depending on different conditions 🧲 
- Steps have a custom lifecycle hooks 🪱
- The code made in this PR doesn't interfere or changes code in components already developed (except `App.vue`) 🏜️ 

## Known limitations

This are some problems and limitations that I believe are not worth solving because they are time consuming and/or only happens in very specific scenarios.

- 😨 Sometimes, arrow tooltip is hidden and I don't know why. I have seen happening only in Brave in android and in Chrome in MacOS. I would say this is more common than it should be, but I have no clue how to solve it.
- 😱 If someone takes the onboarding tour without any transaction in their address, the tour forces the user to click on "Receive free NIM" to continue with the tour. This could be a problem, because we trust that the button will always give the user some NIM. If user is not able to get free NIM from faucet then we will allow user to go to next step but skipping the 'First transaction' step. All logic related to this issue can be improved if we integrate the Faucet into the wallet like we have in `wallet.nimiq-testnet`. **This shouldn't happen anymore, but it has not been tested**
- Legacy accounts will not have access to onboarding tour.
- Network tour is not optimized for mobile phones in landscape view. This is due in part, as the network page is not very good supported.
- Tab navigation with keyboard is quite poor. I didn't find a good way to disable tab navigation if the element has `pointer-events: none`
- Our `translation extraction script` have some issues if we try to translate strings from `.ts` files, a workaround can be seen [here](https://github.com/nimiq/wallet/blob/7d2bc2be4afaf62b5d28e0e4aa9f856f019101b1/src/lib/tour/network/NetworkTourTexts.ts#L3)
- If user resizes during the tour to a small screen, it might happen that once he ends the tour, the buttons in the wallet don't react to events like on-click. This only happens in Chromium-based browsers, and not always! The probability of this is rare, that's why we won't fix it. To see this error, you MUST resize the window and not using "Device Tool Bar"

## Technical explication
- New components: `Tour.vue` and `TourLargeScreenManager.vue`. New welcome modal named: `DiscoverTheWalletModal.vue`  
- `/lib/tour`: Here are the [`types`](https://github.com/nimiq/wallet/blob/onmax/tour/src/lib/tour/types.ts) for the Tour and steps for each tour have been defined. A tour basically is composed by a series of steps. For now, only two tours have been developed: [`onboarding`](https://github.com/nimiq/wallet/tree/onmax/tour/src/lib/tour/onboarding) and [`network`](https://github.com/nimiq/wallet/tree/onmax/tour/src/lib/tour/network). Each step has:
  - `tooltip`: This is the configuration for the tooltip. Basically, it is using the API from [Vue Tour](https://github.com/pulsardev/vue-tour) under the hood with the only change that `content` MUST be an array of strings
  - `ui`: Which has:
    -  `fadedElements`: Elements that we need to add `opacity`. This elements will not have interactivity. See "Disabling interactivity and adding opacity" section.
    - `disabledElements`: Elements that we need to remove interactivity.  See "Disabling interactivity and adding opacity" section.
    - `isNextStepDisabled`
  - `lifecycle`: Custom functions that the developer can execute in different phases of the transition between steps. See "Steps lifecycle" section

#### Disabling interactivity and adding opacity
We are using `data-attribute` instead of `CSS classes` because it is easier to track which elements should be updated  as `data attributes` allow to use a selector like `[data-opaified="1"]` and also selectors like `[data-opacified]`. See `Tour.vue@_removeAttributes`

#### Communication between  `Tour.vue` and `TourLargeScreenManager.vue`
As these two component don't have any close relationship but the need to send data/actions to each other, they are using root instance to emit events. For example [here](https://github.com/nimiq/wallet/blob/onmax/tour/src/components/Tour.vue#L331)

#### Getting data from other component
The idea is that the Tour function works as explicitly as possible, therefore, changing code of other original components have been avoided but sometimes the Tour logic needs a component's instance, so this function have been added: 
[searchComponentByName](https://github.com/nimiq/wallet/blob/35c1b324632ac507a00c582d6b94e0b46c1bea99/src/lib/tour/index.ts#L21)

#### Steps lifecycle
See `Tour.vue@_moveToFutureStep()`
```
            │   Compute future index using: goToNextStep() && goToPrevStep() which are the entrypoints
            │   tour.stop()
            ▼
 ┌───────────────────────────────┐
 │         unmounted()           │   - Old step unmounting function
 └──────────┬────────────────────┘  
 ┌───────────────────────────────┐
 │          created()            │   
 └──────────┬────────────────────┘
            │  change the path if neccessary 
            │  Update UI (fadedElements, non-interactable buttons/divs...) 
            ▼
┌────────────────────────────────┐
│          mounted()             │   - Returns the unmounted() function
└────────────────────────────────┘  
            │ tour.start()
            │
            ▼
WAITING FOR USER INTERACTION
FOR NEXT ACTION
```

## New dependencies
- [Vue Tour](https://github.com/pulsardev/vue-tour) which uses Vue 2. 
_In the future we could use [Vue 3 Tour](https://github.com/alexandreDavid/vue3-tour)_
- Vue Tour uses [Popper JS](https://popper.js.org/)

## How it has been tested?

- Tested **heavily** in Windows 11: firefox, brave and edge
- Tested in Android: Chrome and brave
- Tested in iPhone 13


src/App.vue
```diff
@@ -1,5 +1,11 @@
 <template>
     <div id="app" :class="{'value-masked': amountsHidden}">
+        <!-- (?) This could be moved to Groundfloor.vue -->
```
onmax: What are you thoughts about where the `Tour` component should be placed?

src/components/BtcTransactionList.vue
```diff
@@ -62,7 +62,7 @@
         </RecycleScroller>
 
         <div v-else-if="!searchString" class="empty-state flex-column">
-            <h2 class="nq-h1">{{ $t('Your transactions will appear here') }}</h2>
+            <h2 class="nq-h1">{{ $t('Nothing here yet') }}</h2>
```
onmax: I am not sure if this PR should include this change. In Figma there are a few different texts, should I update them?

src/components/TransactionList.vue
```diff
@@ -61,15 +61,17 @@
                     >
                         <a slot="cashlink" href="#cashlink" @click.prevent="onCreateCashlink">{{ $t('Cashlink') }}</a>
                     </i18n>
-                    <router-link to="buy" class="nq-button light-blue">
-                        {{ $t('Buy NIM') }}
+                    <router-link to="buy">
+                        <button class="nq-button light-blue">
```
onmax, line 65: At some points I need to add `disable` attribute because it changes the UI. In our design system, `a:href` elements don't have a `disabled` option, therefore I have encapsulate the `button` in the `a` element

src/components/icons/GreenNimiqLogoOutlineWithStars.vue
```diff
@@ -0,0 +1,8 @@
+<template>
```
onmax: Not sure if the name of file is great 🙃

src/components/modals/Modal.vue
```diff
@@ -55,8 +55,14 @@ export default defineComponent({
             type: Boolean,
             default: true,
         },
+        showCloseIcon: {
```
onmax: The new `WelcomeModal` don't allow to close the modal

src/components/modals/StartOnBoardingTourModal.vue
```diff
@@ -0,0 +1,96 @@
+<template>
```
onmax: Super basic for now. Needs a ton of work yet

src/composables/useTour.ts
```diff
@@ -0,0 +1,384 @@
+import { Transaction, useTransactionsStore } from '@/stores/Transactions';
```
onmax: In this file all the types and all the tour configurations will be placed. 

I am not sure if this is a good use for a composable though, should I place it somewhere else?

src/composables/useTour.ts
```diff
@@ -0,0 +1,384 @@
...
+}
+
+export type TourStepIndex = MobileOnboardingTourStep | NetworkTourStep;
+
+// eslint-disable-next-line max-len
+// https://github.com/floating-ui/floating-ui/blob/59d15d4f81a2d71d9b42d836e47d7114bc32f7f2/packages/core/src/types.ts#L4
+type Alignment = 'start' | 'end';
+type BasePlacement = 'top' | 'right' | 'bottom' | 'left';
+type AlignedPlacement = `${BasePlacement}-${Alignment}`;
+export type Placement = BasePlacement | AlignedPlacement;
```
onmax: I have been not able to import these types from propper.js

src/composables/useTour.ts
```diff
@@ -0,0 +1,384 @@
...
+        prepareDOMPrevPage?: () => Promise<void>,
+        prepareDOMNextPage?: () => Promise<void>,
+        onBeforeMountedNextStep?: () => Promise<void>,
+        onBeforeMountedPrevStep?: () => Promise<void>,
+        onMountedStep?: (cbu: () => void) => void,
+    };
+
+    ui: {
+        // Elements that must have opacity to focus attention in other elements in screen
+        elementsWithOpacity?: string[], // array of selectors
```
onmax: Not sure if `elementsWithOpacity` and `elementsWithoutInteractivity` are the best names. Suggestions for different names in all the PR are welcome! 🥲

src/composables/useTour.ts
```diff
@@ -0,0 +1,384 @@
...
+                ],
+                params: {
+                    placement: 'bottom-start',
+                },
+            },
+            lifecycle: {
+                onMountedStep: (cbu: () => void) => {
+                    const addressButton = document.querySelector('.address-list > .address-button');
+
+                    // eslint-disable-next-line no-unused-expressions
```
onmax: I had to disable this alert in different locations, do you know how to avoid it?

onmax (commented):
Just some questions I would like to solve 😶‍🌫️

src/components/BtcTransactionList.vue
sisou: If the text has nothing to do directly with the tour, then I would not update them. As you said, there are differences between designs, depending on when they have been branched off the main design.

src/components/Tour.vue
```diff
@@ -0,0 +1,474 @@
...
+                                :key="i"
+                                v-html="$t(content)"
+                            ></p>
+                            <!-- TODO REMOVE ME -->
+                            <div class="remove_me" v-if="currentStep === 1" @click="simulate()">
+                                Simulate Receive NIM
+                            </div>
+                        </div>
+                        <div slot="actions" class="actions">
+                            <button @click="tour.previousStep" v-if="!isMobile">
```
sisou: Why `v-if="!isMobile"` for the prev button? Shouldn't this rather be `!isFirst`?

src/components/TransactionList.vue
sisou: Hmm. Not nice, but I understand the reasoning. Since the `disabled` attribute is only used for styling here (as the link is disabled with the pointer-events:none rule), it makes more sense to add the styling to the style framework maybe?

src/components/icons/GreenNimiqLogoOutlineWithStars.vue
sisou: The file name is perfect! :D

Remember to pass SVGs exported from Figma through SVG-Optimizer (https://jakearchibald.github.io/svgomg/) to make them as small as possible without changing how it looks (play with the decimal slider).

sisou: Also consider to change the `fill` color of the SVG to `currentColor`, so the color can be set as `color` from CSS.

src/components/modals/Modal.vue
sisou: Petition to rename to `showCloseButton`.

src/components/modals/Modal.vue
```diff
@@ -55,8 +55,14 @@ export default defineComponent({
...
+            type: Boolean,
+            default: true,
+        },
     },
     setup(props, context) {
+        const { showCloseIcon } = props;
+
```
sisou: Not used?

src/components/modals/StartOnBoardingTourModal.vue
```diff
@@ -0,0 +1,96 @@
...
+            </div>
+
+            <div class="call-to-action">
+                <button class="nq-button light-blue" @click="startTour">{{ $t('Start Wallet tour') }}</button>
+                <!-- TODO It is an a element? -->
+                <a href="http://">
+                    <span>Skip tour</span>
+                    <!-- TODO Add icon -->
+                    <span>></span>
+                </a>
```
sisou: I believe there is a `.skip` class rule somewhere.

src/components/modals/StartOnBoardingTourModal.vue
```diff
@@ -0,0 +1,96 @@
...
+export default defineComponent({
+    setup(props, context) {
+        function startTour() {
+            context.root.$router.push({
+                name: 'root',
+                params: {
+                    // It has to be a string since it is a value encapsulated in Location.params
+                    // which is Dictionary<string>. Using a 'false' value will lead to the same
+                    // behaviour
+                    doOnboardingTour: 'true',
```
sisou: Maybe instead of just checking if the param is set, its value can be checked to equal the string `'true'`?

In any way, I believe this is not used like this anymore? The tour is started via setting `tour: <name>` in the AccountStore instead?

src/composables/useTour.ts
sisou: `fadedElements` and `disabledElements` maybe? Just a suggestion.

src/composables/useTour.ts
```diff
@@ -0,0 +1,393 @@
...
+    };
+
+    ui: {
+        // Elements that must have opacity to focus attention in other elements in screen
+        elementsWithOpacity?: string[], // array of selectors
+
+        // Elements that shouldn't allow interactivity
+        elementsWithoutInteractivity?: string[], // array of selectors
+
+        disabledNextStep?: boolean,
```
sisou:
```suggestion
        isNextStepDisabled?: boolean,
```

src/composables/useTour.ts
sisou: I believe it's because you use the conditional modifier `?`. Instead you could use the forced modifier `!`, that should make ESLint shut up. But I don't actually get an eslint error when I remove the comment. Maybe not a problem anymore?

src/composables/useTour.ts
```diff
@@ -0,0 +1,393 @@
...
+            },
+            lifecycle: {
+                prepareDOMNextPage: async () => {
+                    if (root.$route.path === '/') {
+                        const addressButton = document
+                            .querySelector('.address-list > .address-button') as HTMLButtonElement;
+                        addressButton.click();
+                    }
+                },
+                onMountedStep: (cbu: () => void) => {
```
sisou: What does `cbu` stand for?

src/composables/useTour.ts
```diff
@@ -0,0 +1,393 @@
...
+                    const addressButton = document.querySelector('.address-list > .address-button');
+
+                    // eslint-disable-next-line no-unused-expressions
+                    addressButton?.addEventListener('click', () => cbu(), { once: true });
+                    // TODO Remove listener somehow if user clicks on the '>' button instead on the address item
```
sisou: A method I learned from Svelte was that the `onMounted` function can _return_ a callback that is run when the component _unmounts_. That could work here too, as long as that callback (if given) is guaranteed to be run. 

**Edit:** This mixes well with [advice further down](https://github.com/nimiq/wallet/pull/76#discussion_r789650046), that all steps should have a `mounted/loaded/setup` and `unmounted/unloaded/teardown` method.

src/composables/useTour.ts
```diff
@@ -0,0 +1,393 @@
...
+                ],
+            },
+        },
+        [MobileOnboardingTourStep.WALLET_BALANCE]: {
+            path: '/',
+            tooltip: {
+                target: '.account-overview .account-balance-container .amount',
+                content: [
+                    'Check the bar-chart to see how your addresses compose your total balance.',
+                    'Currently you have 100% NIM, and no BTC.',
```
sisou: This percentage should probably be dynamic to the actual value shown.

src/composables/useTour.ts
```diff
@@ -0,0 +1,393 @@
...
+                    'Tap on the menu icon to access your account and wallet settings.',
+                ],
+                params: {
+                    placement: 'bottom-start',
+                },
+            },
+            lifecycle: {
+                prepareDOMNextPage: async () => {
+                    await openAccountOptionsModal();
+                },
```
sisou: What you are doing here, I believe doesn't scale well, e.g. when random steps may be skipped because of user state. Instead of having the previous and following steps be responsible for preparing the state for the next step, the step itself should be in charge of preparing the DOM as it needs it. So basically every step should have a `mounted` and `unmounted` callback (or called `loaded` and `unloaded`, you get the idea), in which (for this example) the account menu is opened and closed. Not the previous step be responsible for opening it.

src/composables/useTour.ts
```diff
@@ -0,0 +1,393 @@
...
+            ui: {
+                elementsWithoutInteractivity: [
+                    '.column-sidebar',
+                ],
+            },
+            lifecycle: {
+                prepareDOMPrevPage: async () => {
+                    await closeAccountOptionsModal();
+
+                    root.$router.push('/');
```
sisou: The previous step already has its `path` defined as `'/'`, so that should be automatic. Why do you need to manually navigate here?

src/composables/useTour.ts
```diff
@@ -0,0 +1,393 @@
...
+                        content: ['This is where all your transactions will appear.'],
+                        params: {
+                            placement: 'bottom',
+                        },
+                    };
+                    steps[MobileOnboardingTourStep.TRANSACTIONS_LIST].ui.disabledNextStep = false;
+                }
+            });
+        },
+    };
```
sisou: Why is this outside/after the above step definition?

src/composables/useTour.ts
sisou: I think it's fine. Would also work as a `src/lib/Tour.ts` file.

src/stores/Account.ts
```diff
@@ -104,5 +107,11 @@ export const useAccountStore = createStore({
         setActiveCurrency(currency: CryptoCurrency) {
             this.state.activeCurrency = currency;
         },
+        setTour(tour: AccountState['tour']) {
+            this.state.tour = tour;
+        },
+        removeTour() {
+            this.state.tour = null;
+        },
```
sisou: I don't think this method is necessary, plus its naming is not optimal. What prevents you from using `setTour(null)` wherever you currently use `removeTour()`?

sisou (commented):
Generally nice! :+1:

I was a bit confused when reading through Tour.vue, but I think that is partly because of the functions and variables not being named optimally, so I think there's some potential to improve.

My advice is also in regards to this being "mobile-only". Don't push desktop size too far back, it might get hard to find where to split functionality if done too late.

src/components/TransactionList.vue
onmax: Yeah, I think makes more sense adding `a:disabled` in the style framework, should I do a PR?

src/components/modals/StartOnBoardingTourModal.vue
onmax: Yes, I forgot to remove this. We will be using the AccountStore as you mentioned it. This code therefore will be removed

src/composables/useTour.ts
onmax: Actually is `callback URL` from `JWT` standard which here doesn't make any sense, I will probably rename it to `nextStep` or something similar

src/composables/useTour.ts
onmax: Totally

src/composables/useTour.ts
onmax: I thought that you couldn't reference variable `step` inside of its own declaration like: 

```
const step = {
   fn: () => console.log(step)
}
```

But I was wrong. You can't do that when you are declaring a property of a class but here is not a problem

src/components/icons/GreenNimiqLogoOutlineWithStars.vue
onmax: The color `#21BCA5` is the green color of the logo, should I create a CSS variable for that or hardcode it is fine?

sisou: That is `var(--nimiq-green)`.

src/components/Tour.vue
onmax: This should be `!isMobile && !isFirst`. In the mobile design, the prev button is located in the bar on the bottom of the screen, while in larger screens, this bar on the bottom is not shown

src/composables/useTour.ts
onmax: Yes, this makes a tone of sense. I will change it to `mounted` and `unmounted`

onmax:
This PR is mostly done. Although there are still some tasks that need to be finished:
- Review texts and wording
- Finish new `WelcomeModal` design (currently blocked by overnice)
- Check onboarding tour steps as there are some cases that might change (currently blocked by overnice)

onmax:
Just realised a possible problem:
- 😱 If someone takes the tour without any transaction in their address, the tour forces the user to click on "Receive free NIM" to continue with the tour. This could be a problem, because we trust that the button will always give the user some NIM. But if this does not happen, then the user would get stuck in the tour and would not be able to finish it.

A possible solution could be to skip steps 2 (internally called `TRANSACTION_LIST`) and 3 (aka `FIRST_TRANSACTION`) and go straight to step 4 where the user is introduced to the `bitcoin` address. 

Jumping from step 2 to 4 may be a bit out of place and not giving any feedback to the user may be strange, but perhaps not so much of a problem as the number of times this happens is low.

onmax:
> Just realised a possible problem:
> 
> * 😱 If someone takes the tour without any transaction in their address, the tour forces the user to click on "Receive free NIM" to continue with the tour. This could be a problem, because we trust that the button will always give the user some NIM. But if this does not happen, then the user would get stuck in the tour and would not be able to finish it.
> 
> A possible solution could be to skip steps 2 (internally called `TRANSACTION_LIST`) and 3 (aka `FIRST_TRANSACTION`) and go straight to step 4 where the user is introduced to the `bitcoin` address.
> 
> Jumping from step 2 to 4 may be a bit out of place and not giving any feedback to the user may be strange, but perhaps not so much of a problem as the number of times this happens is low.

You can see 35e8301

onmax:
Onboarding tour is only meant for users who don't have a ledger.

Therefore, in /welcome, the modal should be different if the user has a ledger.

henrysmit420:
Mine bitcoin for free here https://www.btcs.love/invite/4nc58

PR #81, opened 2022-02-12
Improved UX in swap when user only has NIM or BTC

onmax:
## Problem

When user wants to swap assets but only owns one of them, inputs can be confusing as they show that user can put whether positive or negative values. 

This is not true, as if I have for example 1000 NIM and 0 BTC, then I cannot do:
a. +1000 NIM 
b. -0.1 BTC

The expected behaviour in this case would be to **only** allow negative values in the NIM input and positive values in BTC.

This also happens the if the situation is that user owns BTC but no NIM.

## Solution

- Updated UI, so it shows what actions the user can take. Therefore, a new function has been implemented called `getPlaceholder()`
- Updated UX, so even though the user tries to put an invalid value in the inputs, we correct it for him/her

#### User only has BTC
only-btc

#### User only has NIM
only-nim

## Updated AmountInput

I realised that we have two versions of `AmountInput`. I updated only the one that comes with this project as it was more convenient but should I update also `AmountInput` from `vue-component?`

### DIFF
--- a/src/components/AmountInput.vue
+++ b/src/components/AmountInput.vue
@@ -8,8 +8,10 @@
             <input type="text" inputmode="decimal" class="nq-input" :class="{ vanishing }"
                 :placeholder="placeholder"
                 :style="{width: `${width}px`, fontSize: `${fontSize}rem`}"
-                :value="liveValue" @input="onInput"
-                @focus="isFocussed = true" @blur="isFocussed = false"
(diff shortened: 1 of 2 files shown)


sisou (approved):
Very nice, well solved!

I would say to _not_ update the AmountInput component in vue-component, as this functionality is not used anywhere where that component is used.

PR #83, opened 2022-02-14
Center app columns on very large screens

onmax:
## Changes
In order to achieve centering of the app and have the animation at the same time, some elements have absolute positioning and some complex calculations are being made. Hope it is clear.

I don't know how we could achieve the same effect with less complexity.

I have tried to minimise changes to devices smaller than 2000px to reduce the likelihood of breaking something.

#### Test

This PR has been visually tested in:
- latest Chromium version in PC (Edge, Chrome and Brave) -> all screen sizes: mobile, tablet, pc and >2000px
- latest Firefox version in PC -> all screen sizes: mobile, tablet, pc and >2000px
- latest Safari version in iPhone 13
- latest version in Safari in Mac
- latest Chromium version (Brave and Chrome)

### Before:
image

### After
image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -163,6 +163,7 @@ export default defineComponent({
     /* Default: >= 1500px */
     --sidebar-width: 24rem;
     --account-column-width: 70rem;
+    --settings-width: 131rem;
     --address-column-width: 150rem;
 
     @media (max-width: 1499px) {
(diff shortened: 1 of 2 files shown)


sisou:
Please try navigating to and from the network page, and to and from the settings page. It doesn't look right.

sisou (changes requested):
This doesn't work for the transition to and from the network and setting views.

onmax:
I have tried another approach. I'm not a big fan of using fixed distances, but in this case I don't know what other solution we can apply **without changing the DOM**. Also, with this solution, `settings` page is not in the center and it is a bit off

sisou:
This still doesn't work when navigating to the network view. I don't think this can be achieved without rethinking the DOM layout.

https://user-images.githubusercontent.com/1828163/154460731-ecfcf37a-4ca3-4813-b6bb-afe3715b3cba.mp4

sisou (approved):
Amazing, I love it! :heart:

PR #84, opened 2022-02-14
Updated BalanceDistribution NIM addresses tooltips

onmax:
## Distribution NIM bar tooltips

_This is a copy and paste from the original task_
- [ ]  Triangle alignment: triangles should either be centered, or aligned to the whole width of the bar, visually
_Not exactly sure what this mean..._
- [X]  Trigger area: Increase Distribution bar tooltips trigger area. Right now it's just the 4px bar itself, which makes it hard to get to the tooltip. Let's increase the trigger area vertically – and, esp. for small accounts, also fill the gaps between bars.
- [X]  Introduce hover state for bars (decrease opacity for non-hovered bars on hover, see examples below)
- [X]  Invert color to be consistent with the rest of the tooltips

## Before
image

## After
image

## Things to be done

- The first task: "Triangle alignment: triangles should either be centered, or aligned to the whole width of the bar, visually" has not been developed as I am not sure what _aligned to the whole width of the bar_ means
- The opacity for fading the bar not being hovered is 0.5, but I don't know if that's the right value

### DIFF
--- a/src/components/BalanceDistribution.vue
+++ b/src/components/BalanceDistribution.vue
@@ -9,6 +9,7 @@
                         : account.percentage
                     ) * 100 + '%'}">
                     <Tooltip
+                        class="nim"
                         preferredPosition="top right"
                         :container="$el ? {$el: $el.parentNode.parentNode} : undefined"
                         @show="highlightBar(i)"
@@ -48,7 +49,7 @@
                 value-mask/>
         </div>
         <div v-if="hasBitcoinAddresses" class="exchange" ref="$exchange">
-            <Tooltip preferredPosition="top right" :disabled="hasActiveSwap" ref="swapTooltip" noFocus>
+            <Tooltip class="btc" preferredPosition="top right" :disabled="hasActiveSwap" ref="swapTooltip" noFocus>
                 <button
                     :disabled="!totalFiatAccountBalance || hasActiveSwap || !canUseSwaps"
                     @focus.stop="$refs.swapTooltip.show()"
@@ -325,6 +326,10 @@ export default defineComponent({
                     display: flex;
                     height: 8px;
                     padding: 0 0.25rem;
+                }
+
+                &.nim ::v-deep .trigger::after {
+                    background: #1F2046; // Color at the bottom left of the tooltip
                 }
 
                 ::v-deep .tooltip-box {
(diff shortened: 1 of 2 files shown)


src/components/BalanceDistribution.vue
```diff
@@ -292,16 +322,9 @@ export default defineComponent({
...
                 ::v-deep .trigger {
-                    display: block;
-                }
-
-                ::v-deep .trigger::after {
-                    background: white;
-                }
-
-                ::v-deep .tooltip-box {
-                    background: white;
```
sisou: Personally, I liked `white` background better. I think it fits better visually. Plus, the dark background clashes with dark-blue identicons: 
Screenshot from 2022-02-15 10-43-42
@julianbauer What do you think?

sisou (commented):
Great solution! I added a commit to simply the `.bar` selection, turns out we can use the `.bar` elements directly as refs.

onmax:
> Great solution! I added a commit to simply the `.bar` selection, turns out we can use the `.bar` elements directly as refs.

Makes more sense, thank you

onmax:
@julianbauer 

> Triangle alignment: triangles should either be centered, or aligned to the whole width of the bar, visually

What do you mean with "_whole width of the bar_"?

julianbauer:
@onmax here's an example of the behaviour. The triangle should always be center-aligned with the bar it's attached to. If there's enough space, the tooltip is center aligned on top, otherwise it's aligned with the edge of the "bounding box". We'd like a similar behaviour in e.g. modals.

Group 22418

julianbauer:
@onmax here's an example of the behaviour. The triangle should always be center-aligned with the bar it's attached to. If there's enough space, the tooltip is center aligned on top, otherwise it's aligned with the edge of the "bounding box". We'd like a similar behaviour in e.g. modals.

Frame 1361

julianbauer:
Minor style comments:
- Padding should be 12px instead of 20px
- font-weight should be semibold instead of bold (I believe it's bold in your screenshot)

onmax:
> Minor style comments:
> 
> * Padding should be 12px instead of 20px
> * font-weight should be semibold instead of bold (I believe it's bold in your screenshot)

Updated tooltips with your feedback!

image


> @onmax here's an example of the behaviour. The triangle should always be center-aligned with the bar it's attached to. If there's enough space, the tooltip is center aligned on top, otherwise it's aligned with the edge of the "bounding box". We'd like a similar behaviour in e.g. modals.

Got it. This is not going to be fixed for now.

PR #85, opened 2022-02-15
Updated sendBtcTransaction hub request

onmax:
See [this issue](https://github.com/nimiq/keyguard/issues/445) to get more info

- [Hub PR](https://github.com/nimiq/hub/pull/480)
- [Keyguard PR](https://github.com/nimiq/keyguard/pull/446)
 
> In order to test this PR, make sure to update the `package.json` with the appropriate  link to the `hub/client` lib

> Currently CI is failing, not sure how to solve it 🤔. Probably, we need to update the `hub/client` version first before merging

### DIFF
--- a/src/components/modals/BtcSendModal.vue
+++ b/src/components/modals/BtcSendModal.vue
@@ -190,6 +190,8 @@ export default defineComponent({
         } = useBtcLabelsStore();
         const { state: network$, isFetchingTxHistory } = useBtcNetworkStore();
 
+        const delay = ref(12); // Same value as Fee Selector default value
+
         const recipientWithLabel = ref<{address: string, label: string, type: RecipientType} | null>(null);
 
         function saveRecipientLabel() {
@@ -344,9 +346,12 @@ export default defineComponent({
             amount.value = maxSendableAmount.value;
         }
 
-        function updateFee(newFeePerByte: number) {
+        interface FeeSelectorEvent {fee: number; delay: number}
+        function updateFee({ fee: newFeePerByte, delay: newDelay }: FeeSelectorEvent) {
             const isSendingMax = amount.value === maxSendableAmount.value;
             feePerByte.value = newFeePerByte;
+
+            delay.value = newDelay;
             if (isSendingMax) sendMax();
         }
 
@@ -479,6 +484,10 @@ export default defineComponent({
                             value: requiredInputs.value.changeAmount,
                         },
                     } : {}),
+                    fiatCurrency: fiat$.currency,
+                    fiatRate: exchangeRates.value.btc[fiat$.currency] || 0, // Hub will handle this
+                    delay: delay.value,
+                    feePerByte: feePerByte.value,
                 });
 
                 if (!plainTx) {
--- a/src/components/FeeSelector.vue
+++ b/src/components/FeeSelector.vue
@@ -47,7 +47,7 @@ export default defineComponent({
                 }
             }
 
-            context.emit('fee', fee);
+            context.emit('fee', { fee, delay: delay.value });
         });
 
         const speed = computed(() => {


src/components/modals/BtcSendModal.vue
```diff
@@ -190,6 +190,8 @@ export default defineComponent({
         } = useBtcLabelsStore();
         const { state: network$, isFetchingTxHistory } = useBtcNetworkStore();
 
+        const durationConfirmation = ref(3 * 60 * 60); // Fee selector has 2-4h option by default. Setting it as 3h.
```
sisou: This is actually what one of the comments of Julian in the Figma was about. Since the designs were made, the way of displaying the duration changed from a single number to a range (2 _to_ 4 hours). So this needs to be displayed the same way in the Keyguard as it is in the Wallet. Meaning the `duration` parameter must be a range. Maybe `[from: number, to: number]`?

src/components/modals/BtcSendModal.vue
```diff
@@ -344,9 +346,11 @@ export default defineComponent({
             amount.value = maxSendableAmount.value;
         }
 
-        function updateFee(newFeePerByte: number) {
+        interface FeeSelectorEvent {fee: number; durationConfirmation: number}
+        function updateFee({ fee: newFeePerByte, durationConfirmation: _duration }: FeeSelectorEvent) {
             const isSendingMax = amount.value === maxSendableAmount.value;
             feePerByte.value = newFeePerByte;
+            durationConfirmation.value = _duration;
```
sisou: I'm not a fan of using underscores for shadowed variables. In this case, you can use the same mechanism as the other variable, and call it `newDuration`.

src/components/modals/BtcSendModal.vue
```diff
@@ -479,6 +485,11 @@ export default defineComponent({
                             value: requiredInputs.value.changeAmount,
                         },
                     } : {}),
+                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                    // @ts-ignore
+                    fiatCurrency: fiatCurrency.value,
```
sisou: Instead of the usage of a new variable, just to assign it here, you can use the store here directly, without the getter in between:

```suggestion
                    fiatCurrency: useAccountStore().state.activeCurrency,
```

sisou (changes requested):

src/components/modals/BtcSendModal.vue
onmax: Please see this [new interface](https://github.com/nimiq/keyguard/pull/446/files#diff-a05f2434a935e0ed8773231410caca04d096d291588c831e79aca6e024aaf2fbR180)

Basically: `durationConfirmation` no longer exists and it will be renamed to `delay`, which is a value from 1 to 100 (I think).

Therefore, the `keyguard` will receive the `delay` value and the keyguard will decide what to show depending on that value, which at the moment is `15m`, `2-4h` or `6h+` (see [these lines](https://github.com/nimiq/keyguard/pull/446/files#diff-d4a2fa3753bc27ea59222ffaf34e8e555d4a6a96874837551bbfe3f3165c191cR139-R143)) 

This solution is motivated by what we do with `fiatRate` instead of directly passing the transaction value in fiat. We pass to the keyguard the most basic data, and in the keyguard we transform that data to `15m`, `2-4h` or `6h+`. 

## Pros
- `delay` could be used for other calculations.

## Cons
- we are converting `delay` to `15m`, `2-4h` or `6h+`  twice (one in the wallet and another in the keyguard), and. Harder to maintain.

We could just pass to the keyguard as a string with the estimation duration and not overcomplicate things maybe. 

I don't know if passing an array with `from` and `to` will complicate things in the keyguard, because we would need to code a function that calculates every possible combination like: `30m-2h` or `4-6h+`, and I don't think is doable, not for the code but translations.

Not sure which solution is better. Wdyt?

sisou: I think this is a **great** solution! You basically pass the number of blocks that the user selects as the `delay`, which means the Keyguard does the same calculation on the same basis (number of blocks delay) as the wallet, which is the most consistent! It remains to be seen if we need to rename the variable to `blockDelay`.

For what it's worth, passing a string would be horrible, because it would not allow translations, nor any validation really.

sisou (approved):

PR #90, opened 2022-02-24
Fading transition in groundfloor disabled if network is active

onmax:
## Problem

When going from network to settings, if account overview was active previously, fading transition between 'account' and 'settings' was visible.

https://www.loom.com/share/fc8674f0706544c9895ff3b515ee0439

https://www.loom.com/share/d1ac5375edf04cf28f77db4155e35931

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -426,17 +426,21 @@ body {
     &.slide-right-leave-to {
         transform: translate3d(calc(100vw - var(--sidebar-width)), 0, 0);
     }
-}
 
-.address-overview {
-    &.slide-right-enter,
-    &.slide-right-leave-to {
-        transform: translate3d(calc(100vw - var(--account-column-width) - var(--sidebar-width)), 0, 0);
+    .address-overview {
+        &.slide-right-enter,
+        &.slide-right-leave-to {
+            transform: translate3d(calc(100vw - var(--account-column-width) - var(--sidebar-width)), 0, 0);
+        }
+
+        .slide-right-enter &,
+        .slide-right-leave-to & {
+            transform: translate3d(calc(-1 * var(--account-column-width)), 0, 0);
+        }
     }
 
-    .slide-right-enter &,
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -426,17 +426,21 @@ body {
     &.slide-right-leave-to {
         transform: translate3d(calc(100vw - var(--sidebar-width)), 0, 0);
     }
-}
 
-.address-overview {
-    &.slide-right-enter,
-    &.slide-right-leave-to {
-        transform: translate3d(calc(100vw - var(--account-column-width) - var(--sidebar-width)), 0, 0);
+    .address-overview {
```
onmax, line 430: I decided to reorder these CSS classes to make CSS Specificity higher

src/components/layouts/Groundfloor.vue
```diff
@@ -2,12 +2,12 @@
     <div class="groundfloor flex-row">
 
         <div>
-            <transition name="fade">
+            <transition name="fade" v-if="$route.path !== '/network'">
```
onmax, line 5: This solution might not scale well if we add new pages that do not use groundfloor

sisou: We'll cross that bridge when we get to it. It's unlikely that there will be a new page equivalent to the network. The current way of having the network "behind" the other pages, is to visualize that the network is the foundation of the app (going to the network page is like peeling away the top-layer). So even if e.g. Staking would become it's own main page (it is not currently planned to become), then it also wouldn't be this type of transition.

sisou:
Two things changed between "Before" and "After" that I think you didn't want to:
1. The address list now fades out when going to Network
2. The transaction list now does not longer move to the right on the .groundfloor when going to Network

I have a fix for 1., looking into 2. now.

sisou:
Superseeded by #91.

sisou:
Superseeded by #91.

PR #91, opened 2022-02-25
Fade transition in groundfloor disabled when coming from network

sisou:
Replaces #90.

### DIFF
--- a/src/components/layouts/Groundfloor.vue
+++ b/src/components/layouts/Groundfloor.vue
@@ -2,12 +2,12 @@
     <div class="groundfloor flex-row">
 
         <div>
-            <transition name="fade">
+            <transition :name="preventNextTransition ? 'none' : 'fade'">
                 <keep-alive>
                     <router-view name="accountOverview"/>
                 </keep-alive>
             </transition>
-            <transition name="fade">
+            <transition :name="preventNextTransition ? 'none' : 'fade'">
                 <router-view name="settings"/>
             </transition>
         </div>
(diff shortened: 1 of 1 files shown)


src/components/layouts/Groundfloor.vue
```diff
@@ -25,10 +25,20 @@
...
 
 <script lang="ts">
-import { defineComponent } from '@vue/composition-api';
+import { defineComponent, ref, watch } from '@vue/composition-api';
 
 export default defineComponent({
+    setup(props, context) {
+        const preventNextTransition = ref(false);
 
+        watch(() => context.root.$route, (to, from) => {
```
onmax:
```suggestion
        watch(() => context.root.$route, (to) => {
```

src/components/layouts/Groundfloor.vue
```diff
@@ -61,5 +72,9 @@ export default defineComponent({
 
         transition: opacity var(--transition-time) var(--nimiq-ease);
     }
+
+    .settings {
```
onmax, line 76: Maybe this should be better in App.vue, as I see it, there are two problems:
- `.settings` selector is too general and the box in the sidebar with the `settings` class is also getting this class
- The rest of elements have their dimensions in App.vue, not Groundfloor.

onmax (approved):
It is way cleaner like this, and easier to understand. Thanks for improvising it 😄.

About the changes, I can commit it myself or we can just merge, whatever you prefer

src/components/layouts/Groundfloor.vue
sisou: - The `.settings` selector here is in scoped CSS, so it does not affect the Settings button in the sidebar.
- I tried moving the `.settings` styles from here into `App.vue`, but it somehow got the wrong specificity and ended up preventing the transitions from playing right when going between root and settings.

I will merge this now and release it (as there is another critical fix to be deployed), but feel free to play around with the styles, and commit directly to master if it works.

sisou:
I added you as the co-author of this commit, @onmax.

Issue #93, opened 2022-03-01
 scroll mask overflows in address list


onmax:
**Reported by**: _Community member in a Telegram group_
**Browsers**: Brave Version 1.35.103 Chromium: 98.0.4758.102 (Official Build) (64-bit). Probably this bug also occurs in other browsers.

## Description
Overflow with 'scroll-mask top' & 'scroll-mask bottom' when screen width is between 2000px-2070px

## Example
![image](https://user-images.githubusercontent.com/22072217/156215969-629f3e18-b587-49b7-8c77-9713a7394a55.png)

sisou:
Resolved by #94.

PR #94, opened 2022-03-01
Fixed visual bug with overflown scroll mask

onmax:
Solves #93. 

This solution has hardcoded values which are highly discouraged, but the rest of the layout is also hardcoded, so I though it wouldn't be a problem.

I can look for another solution if necessary. 😄

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -164,7 +164,7 @@ export default defineComponent({
     --sidebar-width: 24rem;
     --account-column-width: 70rem;
     --settings-width: 131rem;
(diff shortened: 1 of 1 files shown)


sisou (approved):
After some experimentation, it turns out the fix is to instead only apply the centering logic for big screens starting at 2070px, as otherwise a gap is created between 2000-2070px on the right of the address column (as the width is lower than the screen width). With this fix, all columns grow at the 2000px breakpoint, while the address column still adjusts to the available screen width (because its width is clamped), until 2070px, where it stays centered.

Issue #95, opened 2022-03-02
Address label flickering


onmax:
**Reported by**: _Nimiq team member_
**Browsers**: Chromium: 99. Probably this bug also occurs in other browsers.

## Description
When user hovers the address label in a computer, he/she experiences a fast flickering as shown here: https://www.loom.com/share/d85bbe4339fa4014ae794ffcc0e60d56

Looking at network requests, you can see that 'Fira Mono Bold' is being requested once the user hovers the address label. The request takes less than a second to be completed, but meanwhile, Chromium sets 'monospace bold' as a fallback and then it changes to 'Fira Mono Bold'. It's the change of font what produces the flickering.

This behaviour, as far as I can tell is different in Firefox, which decides not to change the font at all and that's why you don't see a flickering. Although I am not sure about this.

## Solution
Instead of downloading 'Fira Mono Bold' on demand, just do it ahead-of-time

sisou:
This is rather a case of a missing `font-display: swap` rule, which Google Fonts included, but I have missed in my custom `font.css`:
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

onmax:
reopenned it because I am not sure this is solved. Just checked [testnet wallet](https://wallet.nimiq-testnet.com/) with the commit [383e54e](https://github.com/nimiq/wallet/commit/383e54e3a2607bdfe95ff8c651dd0ca793ade688) and still flickering

julianbauer:
@onmax we could switch to using fira code instead of fira mono (the only difference is if there's a slash in the 0 or a dot and the alignment of some characters) and their variable font, that would solve the issue as a side effect. https://fonts.google.com/specimen/Fira+Code

sisou:
@julianbauer I have pushed a branch `fira-code` that switches to the Fira Code Variable Font, check it out.

I personally don't like the slashed 0. Additionally, parts of the UI that come from the nimiq/vue-components repo still use Fira Mono (that would have to be changed, if we change everything to Fira Code).

julianbauer:
Nice. Just checked it out, looks good! You could now add a font-weight transition to e.g. the copyable address underneath the address name to have a super smooth hover.

I don't have a strong opinion about the slashed vs dotted 0, I slightly prefer the dotted Fira Mono version, but given the benefits we get from a variable font (less file size to download, only one request, font-weight animation), I'd say let's go for fira code

PR #96, opened 2022-03-02
Fix address label flickering on hover

onmax:
Solves #95


onmax:
Splited this PR into 2: #97 and #98

PR #97, opened 2022-03-02
Fix address label flickering on hover 

onmax:
Solves #95 

#### Before hovering
image

#### Hovering - first phase 
image

#### Hovering - second phase
image

You can see the video here: https://www.loom.com/share/5f81a1901b9445428eee628756886efe.

## Explanation

1. User mouse hover the address
2. The browser doesn't have the font with the bold variation downloaded, so it requests the bold variation to wherever we are hosting the font. We know this because:
   - For example, you can see the character '0' in the screenshots. In the first and third you can see that the 0 has a dot. In the second screen shot it has like a '/' character in the middle
   - You can see the network panel. Once you over the address, a request is being made.
3. The default behaviour for Chromium is different to Firefox. I think, Chromium prefers to loads a font that already has a bold variation, whereas Firefox prefers wait to the response with the font and then change the variation, and that's why you don't see flickering in Firefox.

Another solution might be change the default behaviour of Chromium with `font-display`, but I couldn't make it work

### DIFF
--- a/public/fonts/fonts.css
+++ b/public/fonts/fonts.css
@@ -31,6 +31,6 @@
 @font-face {
     font-family: 'Fira Mono';
     font-style: normal;
-    font-weight: 500;
+    font-weight: 500, 700;
     src: local('Fira Mono'), url('FiraMono-Medium.woff') format('woff');
 }


onmax:
Not sure how to fix it with the `font-display: swap` as @sisou mentioned: 

https://github.com/nimiq/wallet/issues/95#issuecomment-1057176820

sisou:
So from your solution here, you think what is going on is that when the user hovers, the CSS `font-weight` is switched to `500`, which makes the browser render a `bold` version of the currently loaded font, until the correct font is loaded and swapped in?

public/fonts/fonts.css
```diff
@@ -30,7 +30,7 @@
 
 @font-face {
     font-family: 'Fira Mono';
-    font-style: normal;
-    font-weight: 500;
+    font-style: normal, bold;
```
sisou: `font-style` is for italics vs normal, not for bold. Bold is the _weight_ and the keyword `bold` is an alias for `700`, thus only works in the `font-weight` rule.

sisou (changes requested):

sisou:
The current change in this PR leads to the semi-bold (500) font being displayed always, even when not hovered.

PR #98, opened 2022-03-02
Removed vue-responsive-components dependency

onmax:
`vue-responsive-components` is only used in one place and at the same time is ~75KB, seems unnecessary for now.

### DIFF
--- a/src/components/layouts/AddressOverview.vue
+++ b/src/components/layouts/AddressOverview.vue
@@ -221,16 +221,19 @@ export default defineComponent({
         const $address = ref<HTMLDivElement>(null);
         const addressMasked = ref<boolean>(false);
 
-        useElementResize($address, maskAddress);
-
         const { isMobile, isFullDesktop } = useWindowSize();
 
-        function maskAddress() {
-            let addressMaskedWidth = 322;
-            if (isFullDesktop.value) addressMaskedWidth = 396;
-            if (!isMobile.value) addressMaskedWidth = 372;
-            addressMasked.value = $address.value!.clientWidth < addressMaskedWidth;
-        }
+        useElementResize($address, () => {
+            let addressWidth: number;
+            if (isMobile.value) {
+                addressWidth = 322;
+            } else if (isFullDesktop.value) {
+                addressWidth = 396;
+            } else {
+                addressWidth = 372; // Tablet
+            }
+            addressMasked.value = $address.value!.clientWidth < addressWidth;
+        });
 
         function hideUnclaimedCashlinkList() {
             showUnclaimedCashlinkList.value = false;
--- a/src/components/layouts/AddressOverview.vue
+++ b/src/components/layouts/AddressOverview.vue
@@ -75,7 +75,7 @@
                         <Copyable v-if="activeCurrency === 'nim'"
                             :text="activeAddressInfo.address" :key="activeAddressInfo.address"
                         >
-                            <div class="address" v-responsive="{'masked': el => el.width < addressMaskedWidth}">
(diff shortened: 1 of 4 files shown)


src/components/layouts/AddressOverview.vue
```diff
@@ -219,6 +217,21 @@ export default defineComponent({
...
+            setTimeout(() => maskAddress(), 1);
+
+            window.addEventListener('resize', maskAddress);
+        });
+        onUnmounted(() => window.removeEventListener('resize', maskAddress));
+
+        function maskAddress() {
+            const addressElement = context.refs.$address as HTMLDivElement;
+            const masked = addressElement.clientWidth < addressMaskedWidth.value;
+            addressElement.classList[masked ? 'add' : 'remove']('masked');
```
sisou: Please don't use this programmatical way of setting classes, when we use Vue. Vue is reactive, you just set a `boolean` variable and specify the class in the template. When doing it the way you have here, it becomes very hard to see where classes come from.

See also the commit that introduced the `vue-responsive-components` in the first place: https://github.com/nimiq/wallet/commit/e3e6f3fd18ba9cb8baae0477f3c8376bd743670f

Plus, there would be a method for this:
```diff
- addressElement.classList[masked ? 'add' : 'remove']('masked');
+ addressElement.classList.toggle('masked', masked);
```

src/components/layouts/AddressOverview.vue
```diff
@@ -219,6 +217,21 @@ export default defineComponent({
...
+            // Using setTimeout just to make sure, this is executed at the end of stack call as sometimes, it adds the
+            // 'masked' class although it's not needed
+            setTimeout(() => maskAddress(), 1);
+
+            window.addEventListener('resize', maskAddress);
+        });
+        onUnmounted(() => window.removeEventListener('resize', maskAddress));
+
+        function maskAddress() {
+            const addressElement = context.refs.$address as HTMLDivElement;
```
sisou: The way to define template refs in the Composition API is with an actual `ref()`:

```js
const $address = ref<HTMLDivElement>(null);
```

Then this becomes:
```suggestion
            const addressElement = $address.value!;
```

src/components/layouts/AddressOverview.vue
```diff
@@ -219,6 +217,21 @@ export default defineComponent({
...
         const showUnclaimedCashlinkList = ref(false);
 
+        onMounted(() => {
+            // Using setTimeout just to make sure, this is executed at the end of stack call as sometimes, it adds the
+            // 'masked' class although it's not needed
+            setTimeout(() => maskAddress(), 1);
+
+            window.addEventListener('resize', maskAddress);
+        });
+        onUnmounted(() => window.removeEventListener('resize', maskAddress));
```
sisou: The original commit (https://github.com/nimiq/wallet/commit/e3e6f3fd18ba9cb8baae0477f3c8376bd743670f) that introduced the `vue-responsive-components` dependency used a `ResizeObserver` on the `$address` element instead. While this here with a window resize listener works, I just want you to be aware that it has worse performance, as it fires on _all_ window resizes, whereas a ResizeObserver would only fire when that window resize actually changes the width of the element.

sisou (changes requested):

src/components/layouts/AddressOverview.vue
onmax: You are completely right. I am going to change the PR now. Huge mistake 😑

onmax:
Given your feedback, I changed the PR. I have added `useElementResize` which only triggers when the element change its size. I am using ResizeObserver which has a 90% market share. So for the 10% remaining I left the `window resize event`.

sisou (approved):

PR #99, opened 2022-03-03
More specific CSS selector for .ticker class

onmax:
## Reproduction
1. Open the Swap Modal and close it
2. Open the ⬆️ Send Modal and close it
3. Open the Swap modal again
image

> This reproduction is not consistent, and it seems random as many times won't happen.

## What I think is going on?
Once we open the Send Modal, it loads its CSS file in a lazy way (that is way you can see it only after opening this modal). This new CSS loaded from `SendModal` also applies to `SwapModal`. Because both rules for the `.ticker` class in `SendModal` and `SwapModal` have the same specificity, the CSS from `SendModal`, which is newer, overwrites the `SwapModal` rules for `.ticker`.

## Possible solutions
1. A good solution for this, I believe it would be `@layer`, but atm, support for this feature is less than 2% in all browsers
2. Making the specificity higher in `SwapModal`. This seems to work, but not sure, as I told in the Reproduction section, this bug is not consistent and it can be happen again in the future. This PR uses this solution.

### DIFF
--- a/src/components/modals/SendModal.vue
+++ b/src/components/modals/SendModal.vue
@@ -5,6 +5,7 @@
         || feeSelectionOpened
         || statusScreenOpened"
         @close-overlay="onCloseOverlay"
+        class="send-modal"
         :class="{'value-masked': amountsHidden}"
         ref="$modal"
     >
@@ -782,6 +783,7 @@ export default defineComponent({
 </script>
 
 <style lang="scss" scoped>
+.send-modal {
     .page {
         flex-grow: 1;
         font-size: var(--body-size);
@@ -1187,4 +1189,5 @@ export default defineComponent({
             border-top-right-radius: 1.75rem;
         }
     }
+}
 </style>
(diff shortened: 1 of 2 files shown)


mraveux (approved):
Looks nice to me. The duplication of `font-size: var(--size); font-weight: bold;` is a bit annoying but we'll go with it since code looks nicer this way.

onmax:
Agree, but it was that or duplicate selectors which imo look worse... Don't know any other way

sisou (approved):
This should work nicely. But according to your description, the `SendModal` styles for ticker still apply for the ticker in the `SwapModal`, they are now simply overwritten with this new specificity.
Can we make the same change in the `SendModal`, so those styles don't get applied to the other modal in the first place?

onmax:
> This should work nicely. But according to your description, the `SendModal` styles for ticker still apply for the ticker in the `SwapModal`, they are now simply overwritten with this new specificity. Can we make the same change in the `SendModal`, so those styles don't get applied to the other modal in the first place?

Yes, definitely. I just added a new commit 4771f46 with that. Although, you can see that `SendModal` CSS is already scoped, so it shouldn't be leaking any CSS, so I would say that 4771f46 doesn't solve anything. 

On the other hand, it has something to do with `AmountInput`  component which is being shared between `SendModal` and `SwapModal`

PR #100, opened 2022-03-03
Address selector in Send Modal overflows

onmax:
image

### DIFF
--- a/src/components/modals/SendModal.vue
+++ b/src/components/modals/SendModal.vue
@@ -797,7 +797,6 @@ export default defineComponent({
         justify-content: space-between;
         align-items: center;
         flex-grow: 1;
-        overflow-y: visible; // needed for ios Safari
         position: relative;
     }
 


onmax:
Not sure what is the purpose of 59ab6f15036c422392303134504202914f3ccc07. @Yukiioz do you remember? I have been testing this now in browser stack, seems ok

mraveux:
I did it for a reason but i don't remember why exactly. I think it was to fix a `flex` bug in ios safari for specific ios versions. Not sure.

onmax:
Mmmm, ok. I don't know what is the right call here. 

I would say that the `flex` bug in iOS is already gone, as it was year and a half year ago. At the same time, this rule is breaking the address selector in every browser and seems that iOS is working just fine, so I would go ahead with this PR, but maybe there is something I am missing 🤔

sisou (changes requested):
This overflow fix was originally introduced because the currency-selector tooltip, which also includes the buttons for "Set fee" and "Send all", was getting cut off on Safari because it flowed outside the modal ([`git blame` is your friend here](https://github.com/nimiq/wallet/blame/b1328777f126b59728ec60c841bb22537bb6d249/src/components/modals/SendModal.vue#L800) :wink:).

Since this overflow rule now creates a problem in an overlay for which it is not needed, how about instead namespacing this rule to the specific `.page-body` that it is meant for.

```diff
    .page-body {
        /* ... */
-       overflow-y: visible; // needed for ios Safari
+       &.page__amount-input {
+           overflow-y: visible; // needed for ios Safari
+       }
    }
```

onmax:
We could do that for sure. But before removing `overflow-y: visible;`, the specific reason of the commit 59ab6f15036c422392303134504202914f3ccc07 was most likely due to a Safari Bug. That was one and half years ago, so I thought that the bug could be already fix. Nonetheless, I tested this PR as well in Safari in MacOS, Ipad and iphone. 

Here are some screenshots:
iphone1
mac1

Also tested in ipad and macos with different screen sizes. 

As you can see, now "Send all" and "Set fee" are inside the modal always, and therefore the problem might be gone. Probably, I am missing something, and I don't have any problem doing you suggestion instead. 

Just wanted to share how I reached this conclusion. Also, I believe is always better to remove code rather than keep legacy code. Any thoughts?

sisou:
Thanks for the testing! I agree with you that removing unneeded code is better.

@Yukiioz do you remember specifically why you added this? If my hunch is correct, and we didn't have scrolling inside the currency selector tooltip back then, then it was most likely the tooltip overflowing the modal, which is not relevant anymore. If you agree, it can be deleted.

mraveux:
I have no idea, I don't remember. 😕 
But if it's now useless, then i'm ok removing it.

sisou (approved):

PR #101, opened 2022-03-04
Improved copyable in ReceiveModal

onmax:
Changed the `:hover` state in `ReceiveModal`

## Before
image

## After
image

### DIFF
--- a/src/components/modals/ReceiveModal.vue
+++ b/src/components/modals/ReceiveModal.vue
@@ -16,9 +16,7 @@
                 <QrCodeIcon/>
             </button>
             <Identicon :address="activeAddressInfo.address"/>
-            <Copyable :text="activeAddressInfo.address">
-                <AddressDisplay :address="activeAddressInfo.address"/>
-            </Copyable>
+            <AddressDisplay :address="activeAddressInfo.address" copyable/>
             <button
                 v-if="activeAddressInfo.type === AddressType.BASIC"
                 @click="receiveLinkOverlayOpened = true"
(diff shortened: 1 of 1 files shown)


sisou:
Was there an issue about this? Did designers say this should be changed? I personally don't like the new darker hover background.

src/components/modals/ReceiveModal.vue
```diff
@@ -127,31 +123,11 @@ export default defineComponent({
...
-            transition: opacity 0.3s var(--nimiq-ease);
-            font-weight: 500;
-            opacity: 1;
-        }
-
-        &:hover,
-        &:focus,
-        &.copied {
-            background: rgba(5, 130, 202, 0.07); // Based on Nimiq Light Blue
-        }
```
sisou, line 146: These styles should be ported, to keep the same hover style.

src/components/modals/ReceiveModal.vue
```diff
@@ -127,31 +123,11 @@ export default defineComponent({
         margin-top: 3rem;
     }
 
-    .copyable {
-        padding: 0.5rem;
+    & ::v-deep .copyable {
         background: var(--nimiq-highlight-bg);
-        border-radius: 0.625rem;
-        margin: 1rem 0 4rem;
```
sisou, line 133: This border-radius and margins should stay, to keep the same style.

sisou (changes requested):

julianbauer:
@sisou the issue was about the text not turning blue on hover, the background should indeed stay in lightblue

PR #103, opened 2022-03-11
Enable Portugal for OASIS

Saiv46:


sisou:
Sadly, this is not how it works. This list of countries is auto-generated from a list of participating banks in the SEPA Instant network **RT1** (https://www.ebaclearing.eu/services/rt1/participants/), the network that OASIS currently supports. Sadly, Portuguese banks appear to be in another SEPA Instant network, which is sadly not compatible: **TIPS** (https://www.ecb.europa.eu/paym/target/tips/facts/html/index.en.html).

When OASIS becomes compatible with TIPS, this list will be updated and Portugal will gain access.

ilyaLibin:
@sisou Hey! The issue that PR was aiming to solve is UX with oasis swap.
The user is trying to convert crypto to fiat, but validation blocks him from doing it because Portugal is not in the SEPA list.
Please note that SEPA is not required for the swap, but there is no way for the user to overcome this validation and proceed with the process. Please suggest what to do.
image

ilyaLibin:
On the forum, people suggested just choosing any bank from the list, but then I get this validation error above.

PR #127, opened 2023-03-14
Update en.po

Eligioo:
Fix typo


sisou (changes requested):
The strings in the en.po file are extracted from the code with the `yarn i18:extract` command. So you need to fix the source string in the code for this change to have any effect 😉

Eligioo:
> The strings in the en.po file are extracted from the code with the `yarn i18:extract` command. So you need to fix the source string in the code for this change to have any effect 😉

Thanks for the info. I've updated the source.

sisou (approved):
Haha, you could have left the en.po file change, too :)

PR #129, opened 2023-08-12
Allow search bar to search for label and extradata

Albermonte:
Previously there was unexpected behavior when searching for address labels and "extra data" was not used for the search in BTC or USDC.

image
image

As you can see, nothing is shown...

Now, you can search for labels or any word in the data and it will be displayed.

image

image
image

### DIFF
--- a/src/components/TransactionList.vue
+++ b/src/components/TransactionList.vue
@@ -90,7 +90,7 @@ import TransactionListItem from '@/components/TransactionListItem.vue';
 import TestnetFaucet from './TestnetFaucet.vue';
 import CrossCloseButton from './CrossCloseButton.vue';
 import { useAddressStore } from '../stores/Address';
-import { useTransactionsStore /* , Transaction */, TransactionState } from '../stores/Transactions';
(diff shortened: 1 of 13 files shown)


src/components/TransactionList.vue
```diff
@@ -99,6 +99,7 @@ import { isProxyData, ProxyType, ProxyTransactionDirection } from '../lib/ProxyD
 import { createCashlink } from '../hub';
 import { useConfig } from '../composables/useConfig';
 import { useWindowSize } from '../composables/useWindowSize';
+import { useTransactionInfo } from '../composables/useTransactionInfo';
```
sisou, line 102: Can these composables also be used in the Transaction modals? `TransactionModal`, `BtcTransactionModal` and `UsdcTransactionModal`. I believe there the same computations are made?

sisou (approved):

src/components/TransactionList.vue
Albermonte: Good catch, I have replaced all the duplicated code. Not as much as in the TransactionList but still a good amount 👍🏼

PR #146, opened 2024-02-21
Fix typos

rex4539:


rex4539:
@sisou Could you please take a look?

sisou (changes requested):
Thanks! Please revert the change in `BitcoinJS.min.js`, that is a generated file from a dependency.

sisou (approved):

PR #159, opened 2024-04-20
Update Prestaking.ts (Rabinovitch)

Shekelme:
Just adding 'Rabinovitch' label to NQ84 S551 N1LB EL22 QHGB S701 T2DR L9A6 N6NL

### DIFF
--- a/src/stores/Prestaking.ts
+++ b/src/stores/Prestaking.ts
@@ -94,6 +94,7 @@ export const usePrestakingStore = createStore({
             },
             'NQ84 S551 N1LB EL22 QHGB S701 T2DR L9A6 N6NL': {
                 address: 'NQ84 S551 N1LB EL22 QHGB S701 T2DR L9A6 N6NL',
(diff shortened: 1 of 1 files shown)


sisou:
Are you planning to make this validator a public pool? Or do you plan other people to delegate stake to it privately?

Shekelme:
As it always seemed to me, validators are designed for people from the outside (besides the owner himself) to delegate to them. So the answer to your question lies in the question itself.
I'm sorry if I didn't understand the question.

sisou:
The answer and question is twofold: every person can delegate their stake to every validator. But the rewards are only paid to the validator. The payout to the stakers is the responsibility of the validator operator.

If you are organizing payouts, you are a pool. But if you don't pay out (like my validator), you are not a pool. Anybody can still stake for my validator, but they won't get rewards for it, only I do.

As a rule of thumb, we only add labels to validators who are pools are going to pay out to their stakers.

PR #160, opened 2024-04-21
Fixed SearchBar width constantly changing in Safari

Albermonte:
Too much JS computation, let CSS work its black magic 
Tweaked max-width duration value to achieve the same animation effect

Reported error:
<video src="https://github.com/nimiq/wallet/assets/14013679/8f98bcaf-480e-4ed0-adf7-b07951bfbbbd" controls="controls" style="max-width: 730px;">
</video>

Fixed:
<video src="https://github.com/nimiq/wallet/assets/14013679/37d603ec-9fa3-4960-90c7-7167f2429589" controls="controls" style="max-width: 730px;">
</video>

Tested Safari and Brave on macOS. Please test on other OS.

### DIFF
--- a/src/components/SearchBar.vue
+++ b/src/components/SearchBar.vue
@@ -261,3 +261,10 @@ input {
     }
 }
 </style>
+
+<style>
+.actions-mobile:has(.search-bar:focus-within)
+.container ~ :is(.reset.icon-button,.prestaking-button,button:has(.cashlink-button)) {
+   display: none;
+}
+</style>
--- a/src/components/SearchBar.vue
+++ b/src/components/SearchBar.vue
@@ -49,7 +49,7 @@ export default defineComponent({
 
         const maxWidth = computed(() => {
             if (!searchBarInput.value) return '100%';
-            return isInputActive.value ? '100%' : '0';
+            return isInputActive.value ? '100%' : 'var(--default-sb-width)';
         });
 
         const handleFocus = () => {
@@ -135,6 +135,7 @@ export default defineComponent({
 
 .search-bar {
     $borderTickness: .1875rem; // 1.5px
+    --default-sb-width: clamp(5.5rem, 100%, 30rem);
 
     position: relative;
     display: flex;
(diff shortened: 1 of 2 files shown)


sisou:
When testing, test also on mobile widths, and with long content there.

onmax:
Seems fine in chromium for Windows.

## Issues 

### Long content 
image

For the long content something like this seems to work:

```css
.cross-close-button {
  right: 1rem;
}

input {
  padding-right: 4rem;
}
```

### Mobile + Staking

image

Maybe hide the staking button and the three dots menu when search bar is focussed?

Maybe too complex CSS 🤔

```css
// src/components/SearchBar.vue - Note that this is another style tag and it is not scoped to the component
<style>
.actions-mobile:has(input.search-bar:focus-within) :is(.reset.icon-button,.prestaking-button) {
   opacity: 0;
}
</style>
```

Albermonte:
Thanks for the feedback @sisou @onmax 

## No more issues with long content
Desktop
image

Mobile
image


## SearchBar on Mobile takes all space
In small phones there was problems when you have the cashlink button + staking + 3 dots, now everything to the right of the searchbar will be set to `display: none` when focused

No focus
image

Focused
image

## Extra
Also made some improvements on how the placeholder text is calculated (tested and working in Safari)

onmax:
Tiny detail:

There is a pixel shift to the top on mobile when you focus the input on mobiles

https://github.com/nimiq/wallet/assets/22072217/d4e1aa7c-bdf6-4406-af6b-60322aee42eb

(Put attention to the to left of the input to see the bug).


I tried to debug it but not sure where the problem is:
- It is not the box-shadow 
- It is not the "cross" button

It is a very tiny thing, so not sure if we want to tackle this now :/

onmax (commented):
LGTM

onmax (approved):
LGTM

mraveux:
Looks good to me. Tested it on Safari and it works as expected. Nice 👍 

The only things that bothered me was the "feel" from the transitions, which I tried to finetune and I think it feels better now.

PR #162, opened 2024-05-22
Prestaking Graph

sisou:
Add PrestakingGraph component, use it in the two relevant places, fetch current global prestake amount from Nimiq.Watch v2 API.

### DIFF
--- a/src/components/prestaking/PrestakingGraph.vue
+++ b/src/components/prestaking/PrestakingGraph.vue
@@ -17,14 +17,22 @@
             ></div>
             <div
                 v-for="i in numberOfInactiveBars" :key="`bar-inactive-${i}`"
-                class="bar nq-blue-bg"
+                class="bar"
                 :class="{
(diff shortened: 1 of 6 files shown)


mraveux:
The graph overall looks really nice 👍 .

I pushed one commit to smooth the animation and add transitions to the "bar" states, and another one with a different "shine" animation that I think looks nice. Up to you to revert the commit back to "pulsing" or not 😄 

The only "issue" that I could find otherwise is on mobile, the graph being aligned to the right and not scaling down make the 3 bars "hide" on the left of the screen. I think either keeping it more centered, or scaling down the graph on small devices would solve the issue. 
localhost_8081_prestaking (1)

sisou:
Nice, well done!

I thought mobile was _good enough_, but it flows outside the viewport indeed on very small screens.

My suggestion is to either
- use CSS to put a negative margin-right on the graph on small screens, but then we have to also move (or remove) the third mark (top right).
- use Javascript to reduce the number of suffix bars on small screens.

PR #163, opened 2024-05-30
Add support for SINPE Movil

onmax:
## What to Review?
- Review the new sidebar logic as the current implementation might not be correct
- Add link to ‘Learn more’ in SinpeModalSellInfoModal.vue
- New `useSinpeMovil` composable with all the logic
- New `AssetTransfer` will all the templating and some logic for swaps that include FIAT. In the future, if we need to implement a swap, we would just need to create a new composable very similar to `useSinpeMovil`
- New Sinpe Mobile Verification
- We need a more generic way to work with the OASIS api lib to allow multiple endpoints. Currently we are setting sinpe movil app-wide but we need to also support Oasis EUR. [See this line](https://github.com/nimiq/wallet/blob/onmax/sinpe/src/main.ts#L85-L86). @sisou is working on a solution
- New `RouteName` enum for routes app-wide to avoid using raw strings. This is not a breaking change. It is more like adding TS to the routes so we have better DX

## TODO

- Merge PR in libswap, oasis-js and fastopot-api. Then update dependencies accordingly in `package.json`
- Merge PR in keyguard and hub
- Remove hardcoded phone number for testgin

## Related PR
- https://github.com/nimiq/keyguard/pull/502
- https://github.com/nimiq/hub/pull/537
- https://github.com/nimiq/libswap-js/pull/1
- https://github.com/nimiq/oasis-api-js/pull/1
- https://github.com/nimiq/fastspot-api/pull/3

## Changelog

- Create new RouteName enum.
- New Sidebar logic to enable/disable "Sell"
- Created new Sell Modal Sinpe Info and Sinpe Mobile Verification Modal.
- Created `AssetTransfer`: A component that will consume a composable that have all the required data for the swap. So UI and logic are divided.
- New components: DualCurrencyInput (will be also ported to Nimiq Pay soon, that's why I took some time to do this now)
- Updated in the whole app the TS code to handle the case of `SwapAsset.CRC` and `SwapAsset.LBTC`, which has been added to the oasis lib
- Created new store for sinpe. We need to keep the phone number, name, tokens created in the phone verification flow and use it in the swap modal
- The swap stores the sms settlement token
-Hub and keyguard updated. Keyguard also includes new designs for Sinpe Movil.
- libswap, oasis-js and fastopot-api updated. See related PR


onmax:
Won't do. Leaving the branch here for future reference.

PR #194, opened 2024-11-20
[WIP] feat: improve validator fetch logic

onmax:
We load the validators api in two occasions:
1. When we run `launchNetwork`
2. In new epochs

---

The webclient when we subscribe to blocks, some of them will be on already ended epochs, so the wallet will update the epoch number two times.

### DIFF
--- a/src/components/icons/Staking/FatSearchIcon.vue
+++ b/src/components/icons/Staking/FatSearchIcon.vue
@@ -1,6 +1,6 @@
 <template functional>
     <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 19 19">
-        <circle cx="8.25" cy="8.25" r="4.5" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
-        <line x1="16.25" y1="16.25" x2="11.75" y2="11.75" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
+        <circle cx="8.25" cy="8.25" r="4.5" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
+        <line x1="16.25" y1="16.25" x2="11.75" y2="11.75" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
     </svg>
 </template>
--- a/src/components/staking/ValidatorIcon.vue
+++ b/src/components/staking/ValidatorIcon.vue
@@ -1,11 +1,12 @@
 <template>
-    <img class="validator-icon" v-if="'logo' in validator" :src="validator.logo" :alt="validator.name" />
+    <img class="validator-icon" v-if="src" :src="src" :alt="validator.name" loading="lazy" />
     <Identicon class="validator-icon" v-else :address="validator.address"/>
 </template>
 
 <script lang="ts">
 import { defineComponent } from '@vue/composition-api';
 import { Identicon } from '@nimiq/vue-components';
+import { useConfig } from '@/composables/useConfig';
 import { Validator } from '../../stores/Staking';
 
 export default defineComponent({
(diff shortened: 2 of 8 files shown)


src/components/staking/ValidatorIcon.vue
```diff
@@ -15,6 +16,14 @@ export default defineComponent({
             required: true,
         },
     },
+    setup(props) {
+        const { config } = useConfig();
+
+        return {
+            src: 'logoPath' in props.validator
+                ? `${config.staking.validatorsEndpoint}}/${props.validator.logoPath}` : undefined,
```
sisou, line 24: Is this deployed to validator-api yet? Or should we keep the fallback of `logo`?

src/network.ts
```diff
@@ -217,7 +217,12 @@ export async function launchNetwork() {
         network$.peerCount = peerCount;
     });
 
+    let lastValidatorRequest = { epoch: Infinity, ts: 0 };
```
sisou: Maybe it's possible to store the current `fetch`'s `AbortController` in this object, too, so we can cancel pending requests when we get a new epoch.

src/network.ts
```diff
@@ -217,7 +217,12 @@ export async function launchNetwork() {
...
     async function updateValidators() {
+        const now = Date.now();
+        if (currentEpoch === lastValidatorRequest.epoch || now - lastValidatorRequest.ts < 60000) return;
```
sisou: Magic number `60000` should be a constant.

src/network.ts
sisou: My suggestion is to type this as
```suggestion
    let currentValidatorRequest: { epoch: number, ts: Date, controller: AbortController } | undefined;
```

Having actual `Date` is much better for debugging.

src/network.ts
```diff
@@ -237,7 +242,8 @@ export async function launchNetwork() {
             payoutType: 'none' | 'direct' | 'restake',
             payoutSchedule: string,
             isMaintainedByNimiq: boolean,
-            icon?: string,
+            logo?: string,
+            logoPath?: string,
             hasDefaultIcon: boolean,
```
sisou: `hasDefaultIcon` should now be `hasDefaultLogo` :D

src/network.ts
```diff
@@ -251,8 +257,10 @@ export async function launchNetwork() {
         };
 
         const { config } = useConfig();
-        const apiValidators = await fetch(config.staking.validatorsEndpoint)
-            .then((res) => res.json()).catch(() => []) as ApiValidator[];
+        const url = `${config.staking.validatorsEndpoint}/api/v1/validators?only-known=false`;
+        const apiValidators = await retry<ApiValidator[]>(
+            () => fetch(url).then((res) => res.json()).catch(() => []), 1000, 3,
+        );
```
sisou, line 280:
```suggestion
        const apiValidators = await retry<ApiValidator[]>(
            () => fetch(url).then((res) => res.json()).catch(() => []),
            1000,
            3,
        );
```

src/components/staking/ValidatorIcon.vue
onmax: This is not deployed yet

src/components/icons/Staking/FatSearchIcon.vue
```diff
@@ -1,6 +1,6 @@
 <template functional>
     <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 19 19">
-        <circle cx="8.25" cy="8.25" r="4.5" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
-        <line x1="16.25" y1="16.25" x2="11.75" y2="11.75" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
+        <circle cx="8.25" cy="8.25" r="4.5" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
+        <line x1="16.25" y1="16.25" x2="11.75" y2="11.75" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="3.0"/>
```
onmax, line 4: we use this component in two places

PR #233, opened 2024-11-25
re-enable network interaction

Albermonte:


danimoh:
Network interaction has been enabled again on master.

PR #235, opened 2024-11-25
Improved Ripple effect

onmax:
- Now the distance between the rings is constant
- Made the SVG smaller
- Simpler CSS


sisou:
@onmax Is this still relevant? I don't think we still have the ripple effect on the icon, do we?

onmax:
I am not sure if this still relevant 

We are still using it in some components https://github.com/search?q=repo%3Animiq%2Fwallet%20HeroIcon&type=code

PR #236, opened 2024-11-25
Add new deploy script

mraveux:
This PR replace the previous release.sh script by a new ./src/deploy.sh script taking care of the release steps for you.

The script is currently untested, thus the PR and not the instant push on master.
Once the script is tested and accepted, the PR will be merged.

Edit:
The script has been tested, and used to deploy the last wallet versions on testnet & mainnet, and seems to work as expected and handle most obvious edge cases.


scripts/deploy.sh
```diff
@@ -0,0 +1,223 @@
...
+    VERSION="$1"
+    shift # Remove version from arguments
+fi
+
+# Default values
+DEPLOYER="matheo"
+EXCLUDE_RELEASE=""
+SYNC_TRANSLATIONS=true
+BUILD_ENV="testnet"
+DEPLOY_ONLY=false
```
sisou: DEPLOYER probably shouldnt' have a default ;)

And I would also like BUILD_ENV not to have a default, but to always require explicitly specifying it on the command line.

scripts/deploy.sh
```diff
@@ -0,0 +1,223 @@
...
+done
+
+# Copy build files
+echo -e "${CYAN}Copying build files...${NC}"
+cp -r ../dist/. dist
+git add dist
+
+# Commit changes
+echo -e "${CYAN}Committing changes...${NC}"
+git commit -m "$(create_message)"
```
sisou: I personally always do a `git status` after adding and before commiting, to check if it doesn't look totally off what I'm doing. E.g. if there's only a minor change in the source, I don't expect all files to have been replaced and vice versa.

So I'd like a confirm step here.

scripts/deploy.sh
mraveux: that's true! I wanted to remove it and forgot, my bad! 👍 

I though about it but went for the default testnet deployment. I'll make it required/explicit then ;)

scripts/deploy.sh
mraveux: Make sense, will add that then 👍

mraveux:
I applied the feedback from @sisou, lemme know if that's better for you ;)

scripts/deploy.sh
```diff
@@ -0,0 +1,274 @@
...
+    VERSION="$1"
+    shift # Remove version from arguments
+fi
+
+# Default values
+DEPLOYER=""
+EXCLUDE_RELEASE=""
+SYNC_TRANSLATIONS=true
+BUILD_ENV=""
+DEPLOY_ONLY=false
```
danimoh, line 58: ~~This overwrites the `DEPLOY_ONLY` set above.~~
You already fixed this.

scripts/deploy.sh
```diff
@@ -0,0 +1,274 @@
...
+
+# Parse remaining arguments
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        --deployer=*)
+            DEPLOYER="${1#*=}"
+            if [ $# -eq 0 ]; then
```
danimoh: You are parsing the `DEPLOYER` from what follows the `=`, so the check for an additional parameter seems wrong.

mraveux:
The script has been tested for release v3.0.11 and seems to work.
I will still add some error handling, in case a command fail so that the script doesn't continue and mess it up.

Except that, the script is considered done, until further feedbacks.

mraveux:
Just pushed some error handling.
Script is considered done for now and ready for final review.

scripts/deploy.sh
mraveux: DEPLOYER is now required, with no default value.

I also added a `--testnet` parameter, and made it required to have at least either `--testnet` or `--mainnet`.

scripts/deploy.sh
```diff
@@ -0,0 +1,292 @@
...
+    fi
+    if [ -z "$BUILD_ENV" ]; then
+        show_usage "Either --mainnet or --testnet must be specified"
+    fi
+fi
+
+# Validate version number format
+if ! [[ $VERSION =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] && [ "$DEPLOY_ONLY" = false ]; then
+    show_usage "Version number must be in format X.Y.Z"
+fi
```
danimoh: This check could move into the `if` block above, which already checks for `[ "$DEPLOY_ONLY" = false ];`

scripts/deploy.sh
```diff
@@ -0,0 +1,292 @@
...
+    echo -e "${BLUE}Connecting to deployment servers...${NC}"
+    for server in "${DEPLOY_SERVERS[@]}"; do
+        echo -e "${CYAN}Connecting to $server...${NC}"
+        if ! ssh "$server"; then
+            echo -e "${RED}Failed to connect to $server${NC}"
+            exit 1
+        fi
+    done
+
+    echo -e "${GREEN}Deployment complete!${NC}"
```
danimoh, line 230: This gives a false sense of safety that everything succeeded.
Maybe print that the user should still look through the logs.

scripts/deploy.sh
```diff
@@ -0,0 +1,292 @@
...
+    exit 1
+fi
+
+# Function to create commit/tag message
+create_message() {
+    echo "Nimiq Wallet v$VERSION
+
+$COMMIT_MSG
+
+$EXCLUDE_RELEASE"
```
danimoh, line 326: I guess this leaves a trailing new line when `$EXCLUDE_RELEASE` is empty?

scripts/deploy.sh
```diff
@@ -0,0 +1,292 @@
...
+echo -e "${BLUE}Checking version against existing tags in deployment repo ($DEPLOY_BRANCH branch)...${NC}"
+EXISTING_DEPLOY_TAGS=$(git tag | grep "^v[0-9].*-$ENV_TAG-" | sed 's/^v\([0-9][^-]*\).*/\1/')
+for tag in $EXISTING_DEPLOY_TAGS; do
+    if ! version_gt "$VERSION" "$tag"; then
+        echo -e "${RED}Error: Version $VERSION is not greater than existing version $tag in deployment repo${NC}"
+        exit 1
+    fi
+done
+
+# Copy build files
```
danimoh, line 372: Should run the `./clean.sh` script here before copying the files.

danimoh (approved):
Very cool, thank you for this :+1:
The issues I showed you in CR, I see you already addressed.
Commenting on some additional things, but you don't necessarily need to address them, if you don't want to.

scripts/deploy.sh
```diff
@@ -57,11 +60,23 @@ DEPLOY_SERVERS=()
...
+    VERSION=""
+    shift
+elif [[ "$1" == "--deploy-only" ]]; then
     VERSION=""
     DEPLOY_ONLY=true
     shift
+elif [[ "$1" =~ ^--.*$ ]]; then
+    # If first argument is any other flag
+    VERSION=""
+    shift
```
danimoh: This discards the flag without any warning. Better would probably be to exit with an error.

scripts/deploy.sh
mraveux: that's true! I removed the unnecessary check ;)

PR #246, opened 2024-12-07
Updated cosigner public keys for testnet

Albermonte:
Mainnet key stays the same

### DIFF
--- a/src/config/config.local.ts
+++ b/src/config/config.local.ts
@@ -163,6 +163,7 @@ export default {
     nimiqPay: {
         cosignerPublicKeys: [
             'c1534c708122968212c6709526801d18218ce9303590f2bca473213d99b24aaf',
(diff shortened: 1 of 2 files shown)


src/config/config.local.ts
```diff
@@ -162,7 +162,7 @@ export default {
 
     nimiqPay: {
         cosignerPublicKeys: [
-            'c1534c708122968212c6709526801d18218ce9303590f2bca473213d99b24aaf',
+            '1a0fc4eb1376b604c7df66d098b29bbb2fcfd68ef328ce98ae0f7d19f4f6d456',
```
sisou, line 166: Please don't _replace_ the key. It's an array for a reason, so we can add new ones when they get rotated.
We still need the old ones around for the historic transactions.

sisou (changes requested):

src/config/config.local.ts
Albermonte: Makes sense, my bad

sisou (approved):
LGTM. Are you allowed to merge?

PR #252, opened 2025-01-21
[WIP] Nimiq Wallet Demo

onmax:
Moved to #263

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ yarn install
 yarn serve
 ```
 
-> You will also need to run the [Nimiq Hub](https://github.com/nimiq/hub#contribute) and [Nimiq Keyguard](https://github.com/nimiq/keyguard/#development) in development, too.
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
     "v-click-outside": "^3.0.1",
     "vue": "^2.6.11",
     "vue-i18n": "^8.15.5",
-    "vue-router": "^3.1.6",
+    "vue-router": "^3.5.0",
```
onmax, line 56: I am working on the wallet playground. I want to add some routes, but since it's only for `playground` mode, I don't want to do it in `router.ts`, but rather dynamically.

I had to update `vue-router` to be able to add `routes` to specific `children` using [`addRoute`](https://v3.router.vuejs.org/api/#router-addroute). Maybe there is another way, but also it doesn't hurt to update now, as it is something less for the future :).

There are no breaking changes between 3.1 and 3.5: https://github.com/vuejs/vue-router/blob/dev/CHANGELOG.md#350-2021-01-25 except for a few TS issues, which you can find in this commit.

@sisou @danimoh do you agree with this decision?

PR #254, opened 2025-01-22
Jeff/misc from sinpe branch

jeffesquivels:
More spin off commits from the sinpe branch.

The ones here are based on da0d25b8c9d8b764b0b439b00a9f20788d908ea4, 8d9379a7b16a484a0a8fd7c7240b32a84cee2e94, 7a109eefb7e46bed6277f3b80318186e956f11ea and 663f5834615fe3b0eb707a4e4c11839b1175da49 without the sinpe movil specific parts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "serve": "yarn pre && yarn vue-cli-service serve --port 8081",
     "build": "yarn pre && yarn vue-cli-service build",
     "lint": "vue-cli-service lint --no-fix",
+    "lint:fix": "vue-cli-service lint",
     "build:bitcoinjs": "yarn --silent browserify bitcoinjs-parts.js -p common-shakeify -s BitcoinJS | yarn terser --compress --mangle --source-map --output public/bitcoin/BitcoinJS.min.js",
     "build:ci": "yarn pre && build=testnet yarn vue-cli-service build",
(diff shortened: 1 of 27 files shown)


sisou:
@onmax Can you explain why the `RouteName` enum is useful? It doesn't improve the situation in this PR I feel.

onmax:
I introduced this because I often misspelled route names in the past. With TS, improve DX in both dev and building. However, I can undo it if you don't think it's worth it.

In code looks like:

```ts
router.push({ name: 'my-modal' }); // before
router.push({ name: RouteName.MyModal }); // after
```

plus, this change does not break anything

PS: Similar principle that you have [commented before](https://github.com/nimiq/wallet/pull/194#discussion_r1850659820) in one of my other PR

sisou:
Well, for now you just moved the string definition from inside the route object to outside. But are the route names also going to be replaced in other places? Only then it makes sense.

onmax:
See #255

jeffesquivels:
Ah sorry, that's probably a mistake on my part, I didn't realized there was a commit where everything else got changed (I thought this was just going to be a partial migration kind of thing and then route names in other places would get changed whenever that code got changed in the future).

sisou (approved):

PR #262, opened 2025-03-10
chore: update vue-router to be able to use `addRoute`

onmax:
I am working on the wallet playground #252. I want to add some routes, but since it's only for playground mode, I don't want to do it in router.ts, but rather dynamically.

I had to update `vue-router` to be able to add routes to specific children using [addRoute](https://v3.router.vuejs.org/api/#router-addroute). Maybe there is another way, but also it doesn't hurt to update now, as it is something less for the future :).

There are no breaking changes between 3.1 and 3.5: [vuejs/vue-router@dev/CHANGELOG.md#350-2021-01-25](https://github.com/vuejs/vue-router/blob/dev/CHANGELOG.md?rgh-link-date=2025-01-22T11%3A33%3A48Z#350-2021-01-25) except for a few TS issues, which you can find in this PR

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10776,10 +10776,10 @@ vue-resize@^0.4.5:
   resolved "https://registry.yarnpkg.com/vue-resize/-/vue-resize-0.4.5.tgz#4777a23042e3c05620d9cbda01c0b3cc5e32dcea"
   integrity sha512-bhP7MlgJQ8TIkZJXAfDf78uJO+mEI3CaLABLjv0WNzr4CcGRGPIAItyWYnP6LsPA4Oq0WE+suidNs6dgpO4RHg==
 
-vue-router@^3.1.6:
-  version "3.4.9"
-  resolved "https://registry.yarnpkg.com/vue-router/-/vue-router-3.4.9.tgz#c016f42030ae2932f14e4748b39a1d9a0e250e66"
-  integrity sha512-CGAKWN44RqXW06oC+u4mPgHLQQi2t6vLD/JbGRDAXm0YpMv0bgpKuU5bBd7AvMgfTz9kXVRIWKHqRwGEb8xFkA==
+vue-router@^3.5.0:
+  version "3.6.5"
+  resolved "https://registry.yarnpkg.com/vue-router/-/vue-router-3.6.5.tgz#95847d52b9a7e3f1361cb605c8e6441f202afad8"
+  integrity sha512-VYXZQLtjuvKxxcshuRAwjHnciqZVoXAjTjcqBTz4rKc8qih9g9pI3hbDjmqXaHdgL3v8pV6P8Z335XvHzESxLQ==
 
 vue-style-loader@^4.1.0, vue-style-loader@^4.1.3:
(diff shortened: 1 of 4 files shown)


sisou (approved):
When you remove the leftover playground import from this PR, and thus the CI will pass, this is good to go! 👍

yarn.lock
```diff
@@ -10776,10 +10776,10 @@ vue-resize@^0.4.5:
   resolved "https://registry.yarnpkg.com/vue-resize/-/vue-resize-0.4.5.tgz#4777a23042e3c05620d9cbda01c0b3cc5e32dcea"
   integrity sha512-bhP7MlgJQ8TIkZJXAfDf78uJO+mEI3CaLABLjv0WNzr4CcGRGPIAItyWYnP6LsPA4Oq0WE+suidNs6dgpO4RHg==
 
-vue-router@^3.1.6:
-  version "3.4.9"
-  resolved "https://registry.yarnpkg.com/vue-router/-/vue-router-3.4.9.tgz#c016f42030ae2932f14e4748b39a1d9a0e250e66"
-  integrity sha512-CGAKWN44RqXW06oC+u4mPgHLQQi2t6vLD/JbGRDAXm0YpMv0bgpKuU5bBd7AvMgfTz9kXVRIWKHqRwGEb8xFkA==
+vue-router@^3.5.0:
+  version "3.6.5"
```
sisou, line 10780: As it's automatically installing the latest version anyway, you can set `3.6.5` as the min version in package.json, too.

sisou: Although _technically_ you are correct that you require only v3.5.0 for the new feature, so it can also stay as is.

onmax: ~~Ok. Then I bumpped to 3.6.5.~~

onmax:
Ready for merge!

yarn.lock
```diff
@@ -9860,7 +9860,7 @@ strict-uri-encode@^1.0.0:
   resolved "https://registry.yarnpkg.com/strict-uri-encode/-/strict-uri-encode-1.1.0.tgz#279b225df1d582b1f54e65addd4352e18faa0713"
   integrity sha512-R3f198pcvnB+5IpnBlRkphuE9n46WyVl8I39W/ZUTZLz4nqSP/oLYUrcnJrw462Ds8he4YKMov2efsTIw1BDGQ==
 
-"string-width-cjs@npm:string-width@^4.2.0", string-width@^4.1.0, string-width@^4.2.0, string-width@^4.2.3:
+"string-width-cjs@npm:string-width@^4.2.0":
```
sisou: Now you introduced changes to some package resolutions that haven't been there before the latest force-push :/

onmax: Not sure why now `yarn` is changing the resolutions and adding all of that noise, I am just running `yarn install`

**I have edited now the `yarn.lock` manually.**

The `yarn.lock` is minimal and I also run `yarn install` to verify that the everything is ok.

sisou (approved):
OK, you can merge now 👍 preferably a rebase merge.

PR #263, opened 2025-03-12
Nimiq Wallet Demo

onmax:
Moved to #278


julianbauer:
Hey Maxi, thanks for putting this together, this will be such a helfpul resource!
I'll comment on your suggested UX improvements, and a bit beyond that on the videos you shared:

julianbauer:
## Buy Fake NIM
Completely agree with you, an interface where users can enter a number and that amount gets added to the balance is perfect. This is not too far from how the original interface for buying via OASIS looked like:
Setup – Select crypto   amount

I think it would be cool to bring it as close as possible to this, as even though users right now will have to go through moonpay et al, this would be the most native way to buy within the wallet. Can we mock a bank and identicon (we do have a generic bank icon too)? Can we show the fiat equivalent or would this be too much work? [Here's the link to the Figma designs](https://www.figma.com/design/D07VHQmmYSaOipkiivRjwv/Wallet-Fiat-Swaps?node-id=6-0&p=f&t=Ruacw6pwydVVUxsh-11)

_Commented by Maxi_:  ✅ Done

Instead of the DEMO banner in the modal, I'd add a note to the whole interface (e.g. where we have the "POS Testnet" note), additionally if we think that's not enough, maybe a little note below the button saying that this is not real money and just for demo purposes.

_Commented by Maxi_:  ✅ Done

Also in the TX list, would be cool to have a [fake bank icon instead of the identicon].(https://www.figma.com/design/D07VHQmmYSaOipkiivRjwv/Wallet-Fiat-Swaps?node-id=756-0&t=Ruacw6pwydVVUxsh-11)

_Commented by Maxi_:  ✅ Done

onmax:
Thanks @julianbauer. You can also add yourself as reviewer and add a review with `request changes` option ;)

> Instead of the DEMO banner in the modal, I'd add a note to the whole interface (e.g. where we have the "POS Testnet" note), additionally if we think that's not enough, maybe a little note below the button saying that this is not real money and just for demo purposes.

Really nice solution. I will update the designs and propose a wording that we like :).

> Also in the TX list, would be cool to have a [fake bank icon instead of the identicon].([figma.com/design/D07VHQmmYSaOipkiivRjwv/Wallet-Fiat-Swaps?node-id=756-0&t=Ruacw6pwydVVUxsh-11](https://www.figma.com/design/D07VHQmmYSaOipkiivRjwv/Wallet-Fiat-Swaps?node-id=756-0&t=Ruacw6pwydVVUxsh-11))

Yep I think we can do that. For that I believe we need to fake the swap as if the user has used supersimpleswap. I didn't want to add a sss swap since the service is offline for more than a year. Should we really add it?

I believe this swap is only in EUR, so a person from other continent don't care about this feature.

julianbauer:
> Yep I think we can do that. For that I believe we need to fake the swap as if the user has used supersimpleswap. I didn't want to add a sss swap since the service is online for more than a year. Should we really add it?

As we're faking it all anyway, could we just... fake it? Like, we probably have a fixed address for this transaction, couldn't we just replace the identicon with the bank icon? Or am I missing sth?

_Commented by Maxi_: ✅ Done

sisou:
I'm against adding a EUR swap, as we are not currently offering that and don't know if we'll ever do again.

julianbauer:
## Stake NIM
Looks good! Probably obvious but 2 things:
- We need some more fake content for the staking pool profile
_Commented by Maxi_: ❌ Should be in the staking PR
- The staking interface in the address overview is still the old one, would be great to work with the new, iterated designs here ([Figma](https://www.figma.com/design/Ibvnl8VXJiRYDdMAShmjNt/Staking-feature?node-id=7506-157122&t=Hzf2EqZLKXs4XGro-11)): 
_Commented by Maxi_: ❌ Should be in the staking PR

Default

julianbauer:
@sisou 
> I'm against adding a EUR swap, as we are not currently offering that and don't know if we'll ever do again.

I'm with you, I wouldn't call it "the EUR swap" or link it to any of the services currently unavailable. It does provide an interface though that I think is nicer and closer to reality than just a single input, and by itself is quite neutral. Considering we have these anyway, why not use them?

_Commented by Maxi_: ✅ Done!

julianbauer:
## Swap NIM
- Probably makes sense to speed this up a little bit – e.g. just stay for 1-2 seconds on each step
_Commented by Maxi_: ✅ Done!
- Before highlighting the BTC tile, first the NIM tile on the left should be filled (to show it's ready)
- You're using the "clap hands" icon for a successful buy experience – for consistency I suggest to use that for the swap as well (or the other way around. I think in the wallet we're using the checkmarks, I do think the hands emoji is nicer though – should then ofc be updated in the wallet too)
_Commented by Maxi_: ✅ Done!
- The "Swap successful" toast stays indefinitely, and clicking on it opens the success step in the modal ;)

julianbauer:
> Reduce the balance to a certain amount. For example, $200 in NIM, $1300 in BTC and a few hundred in USD.
> Dynamic balance depending on your country. We could use an API that gives an economic index based on a country code to set a balance that fits an average person in that country.

Great ideas, agreed!

### Fallback modal
For the fallback modal, we'll come up with a custom design for this. Essentially we want to say "Hey this functionality doesn't exist because this is a demo, but if you're hooked, create a wallet and see the real thing"

_Commented by Maxi_:
<details><summary>✅ Done!</summary>
<p>

image


</p>
</details> 


### Fake addresses
Could we just use special characters in the middle of the addresses? Like NQ26 2Y2M XXXX XXXX XXXX XXXX XXXX MDIF 5SKD? That should generate identicons too, right? On top of that we should definitely deactivate all copy functionality (like clicking the address in Receive modal), and add a note to users to not send funds to these addresses because they don't exist

_Commented by Maxi_:
<details><summary>✅ Done!</summary>

<p>


| NIM | BTC | USDC |
|--------|--------|--------|
| image | image | image |

</p>
</details>

onmax:
> The staking interface in the address overview is still the old one, would be great to work with the new, iterated designs here ([Figma](https://www.figma.com/design/Ibvnl8VXJiRYDdMAShmjNt/Staking-feature?node-id=7506-157122&t=Hzf2EqZLKXs4XGro-11))

Agreed, but outside of the scope of this PR. The UI will be updated automatically when new designs get implemented.

> I'm with you, I wouldn't call it "the EUR swap" or link it to any of the services currently unavailable. It does provide an interface though that I think is nicer and closer to reality than just a single input, and by itself is quite neutral. Considering we have these anyway, why not use them?

I will check how to do it, but I believe I might need to change some code in the application side, which is something I am trying to minimize. I will see how I can do it :)

> Probably makes sense to speed this up a little bit – e.g. just stay for 1-2 seconds on each step

No problem

> Before highlighting the BTC tile, first the NIM tile on the left should be filled (to show it's ready)

Ok, I think I missed a step in the faking process.

> You're using the "clap hands" icon for a successful buy experience – for consistency I suggest to use that for the swap as well (or the other way around. I think in the wallet we're using the checkmarks, I do think the hands emoji is nicer though – should then ofc be updated in the wallet too)

I will update the buy flow with check mark then.

>  Dynamic balance depending on your country. We could use an API that gives an economic index based on a country code to set a balance that fits an average person in that country.

I will see the easiest way to do it without any impact.

> For the fallback modal, we'll come up with a custom design for this. Essentially we want to say "Hey this functionality doesn't exist because this is a demo, but if you're hooked, create a wallet and see the real thing"

Ok, waiting for it.

> Could we just use special characters in the middle of the addresses? Like NQ26 2Y2M XXXX XXXX XXXX XXXX XXXX MDIF 5SKD? That should generate identicons too, right? On top of that we should definitely deactivate all copy functionality (like clicking the address in Receive modal), and add a note to users to not send funds to these addresses because they don't exist

It will not generate identicons as I believe the code will say is a invalid address. but I can look for a work-around.

_Thanks for you valuable feedback. I have updated the TODOs section in this PR to track progress_

julianbauer:
> It will not generate identicons as I believe the code will say is a invalid address. but I can look for a work-around.

Btw I'm basing my assumption on this demo here, where any text creates an identicon. Not sure how applicable this is of course: https://github.com/nimiq/identicons?tab=readme-ov-file

But then, as another workaround, you could use a real address for the identicon, but only show the masked version of it. There's no way to guess the address from the identicon so we should be safe!

onmax:
Under the hood we are using `ValidationUtils.isValidAddress(address)` [here](https://github.com/nimiq/vue-components/blob/master/src/components/Identicon.vue#L19).

But I think is going to change just the text content that we render and leave the identicon itself with a correct address as you mention

julianbauer:
@onmax here's the promised fallback modal design: https://www.figma.com/design/kzdltAuhkXvlM1jLVlSCcU/Nimiq-Wallet?node-id=444-12553&t=P1GfEEvS9FN0wYEQ-11

Demo Fallback modal

onmax:
@julianbauer 

Working on your feedback now.

I just realized that we didn't think about the receive modal. I am thinking of disabling the receive modal completely and always redirecting to the fallback modal. There is no need to show the receive modal at all in the demo. Let me know if this works for you. 

Thanks a lot!

image

julianbauer:
> @julianbauer
> 
> Working on your feedback now.
> 
> I just realized that we didn't think about the receive modal. I am thinking of disabling the receive modal completely and always redirecting to the fallback modal. There is no need to show the receive modal at all in the demo. Let me know if this works for you.
> 
> Thanks a lot!
> 
> image

Ach, I think it's nice seeing how it looks like. If we mask the address properly like in your screenshot, I think there's value in showing it, just for context.

_Commented by Maxi_:  ✅ Done

onmax:
The issue now is the QR codes and request link. How should we handle those?

julianbauer:
I'd suggest to draw the line there: These can be left out :) – show the fallback modal when clicking on them

src/lib/Demo.ts
```diff
@@ -0,0 +1,2011 @@
...
+
+const DemoFallbackModal = () =>
+    import(
+        /* webpackChunkName: 'demo-hub-fallback-modal' */
+        '@/components/modals/demos/DemoModalFallback.vue'
+    );
+
+const DemoPurchaseModal = () =>
+    import(
+        /* webpackChunkName: 'account-menu-modal' */
```
onmax: copy pasta

src/components/modals/demos/DemoModalBuy.vue
```diff
@@ -0,0 +1,333 @@
...
+                    font-size: 2.5rem !important;
+                    padding: 0.375rem 0.75rem;
+                }
+            }
+        }
+    }
+
+    ::v-deep .status-screen {
+        height: 100%;
+  position: absolute;
```
onmax: lint

src/components/modals/demos/DemoModalBuy.vue
```diff
@@ -0,0 +1,333 @@
...
+
+        <div slot="overlay" class="page">
+            <StatusScreen
+                title="Your NIM is on its way!"
+                :state="status"
+                message="This transaction is instant and secure."
+            />
+        </div>
+<!--
+    <PageBody slot="overlay" class="overlay-content">
```
onmax: Remove comment

src/components/layouts/Sidebar.vue
```diff
@@ -359,6 +380,8 @@ export default defineComponent({
                 : null;
         });
 
+        const isDemoEnabled = checkIfDemoIsActive();
```
onmax: isDemoEnabled -> isDemoActive

src/lib/Demo.ts
```diff
@@ -0,0 +1,2011 @@
...
+    daysAgo: number;
+    description: string;
+    recipientLabel?: string;
+}
+
+/**
+ * Defines transaction definitions for demo NIM transactions
+ */
+function defineNimFakeTransactions(): Partial<NimTransaction>[] {
+    const txDefinitions: NimTransactionDefinition[] = [
```
onmax: move txs to the top of file

src/lib/Demo.ts
```diff
@@ -0,0 +1,2011 @@
...
+    incoming: boolean;
+    address: string;
+}
+
+/**
+ * Defines transaction definitions for demo BTC transactions.
+ * Note: We add the "address" field so that incoming transactions use the correct sender address.
+ */
+function defineBtcFakeTransactions(): BtcTransaction[] {
+    const txDefinitions: BtcTransactionDefinition[] = [
```
onmax: move txs to the top of file

src/lib/Demo.ts
```diff
@@ -0,0 +1,2011 @@
...
+    description: string;
+    recipientLabel?: string;
+    incoming: boolean;
+}
+
+/**
+ * Defines transaction definitions for demo USDC transactions
+ */
+function defineUsdcFakeTransactions(): UsdcTransaction[] {
+    const txDefinitions: UsdcTransactionDefinition[] = [
```
onmax: move txs to the top of file

src/lib/Demo.ts
```diff
@@ -0,0 +1,2011 @@
...
+    description: string;
+    recipientLabel?: string;
+    incoming: boolean;
+}
+
+/**
+ * Defines transaction definitions for demo USDT transactions
+ */
+function defineUsdtFakeTransactions(): UsdtTransaction[] {
+    const txDefinitions: UsdtTransactionDefinition[] = [
```
onmax: move txs to the top of file

src/main.ts
```diff
@@ -48,14 +49,22 @@ Vue.use(VuePortal, { name: 'Portal' });
...
     initPwa(); // Must be called as soon as possible to catch early browser events related to PWA
-    await initStorage(); // Must be awaited before starting Vue
-    initTrials(); // Must be called after storage was initialized, can affect Config
-    // Must run after VueCompositionApi has been enabled and after storage was initialized. Could potentially run in
-    // background and in parallel to syncFromHub, but RedirectRpcClient.init does not actually run async code anyways.
-    await initHubApi();
-    syncFromHub(); // Can run parallel to Vue initialization; must be called after storage was initialized.
 
-    serviceWorkerHasUpdate.then((hasUpdate) => useSettingsStore().state.updateAvailable = hasUpdate);
+    const isDemoEnabled = checkIfDemoIsActive();
```
onmax: rename to isDemoActive

onmax (commented):
minor adjustments needed

PR #265, opened 2025-03-17
Replace `.currentRoute` with `useRoute()`

sisou:
https://router.vuejs.org/guide/migration/#The-currentRoute-property-is-now-a-ref-

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -50,7 +50,7 @@ import { nextTick } from '@/lib/nextTick';
 import Sidebar from './components/layouts/Sidebar.vue';
 import SwapNotification from './components/swap/SwapNotification.vue';
 import UpdateNotification from './components/UpdateNotification.vue';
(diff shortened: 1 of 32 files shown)


onmax (approved):
Looks good to me. 

Small detail - https://github.com/nimiq/wallet/blob/useroute/src/components/swap/SwapNotification.vue uses two `useRoute`. 

When we have more than one `useRouter()` we do `const router = useRouter()`. Not sure if we want the same.

sisou:
yes, that is on purpose. The uses are separate enough that I found it reasonable to not have a global `router`, plus I think the `useRouter()` call has basically no overhead, right?

sisou:
Yes, we can be liberal with `useRoute()` I think, as it refers to the current active route, so can change even while the component is mounted, and should thus not be statically assigned on component init.

sisou:
Rebased for merging

PR #270, opened 2025-05-07
Add Matomo (without cookies) for tracking time-to-consensus

sisou:
For testing I set the dev config to the same as the mainnet config, and I see the outgoing event HTTP requests in the network panel (with correct 204 response code), but I cannot see any data in the Matomo dashboard for this site... :/


src/main.ts
```diff
@@ -117,6 +118,10 @@ async function start() {
         initKycConnection();
     });
 
+    if (config.matomo.enabled) {
+        initMatomo(config.matomo.host, config.matomo.siteId, router);
+    }
```
danimoh, line 123: This should be reactive to `config` changes, via a `watch`, as for the other config checks above. This way, tracking can for example be enabled via a trial at runtime.

src/matomo.ts
```diff
@@ -0,0 +1,104 @@
...
+        // Default: 'matomo'
+        // trackerFileName: 'matomo',
+
+        // Overrides the autogenerated tracker endpoint entirely
+        // Default: undefined
+        // trackerUrl: 'https://example.com/whatever/endpoint/you/have',
+
+        // Overrides the autogenerated tracker script path entirely
+        // Default: undefined
+        // trackerScriptUrl: 'https://example.com/whatever/script/path/you/have',
```
danimoh, line 22: By default, `vue-matomo` loads `${host}/${trackerFileName}.js`, e.g. `https://stats.nimiq-network.com/matomo.js`. I.e. code from an external domain is embedded into the Wallet, even without integrity hash, with the associated security risks and increased attack surface.
If to be used, the matomo script should become part of the Wallet codebase, and configured here accordingly.

src/matomo.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        // Enables automatically registering pageviews on the router
+        router,
```
danimoh, line 25: This results in personal data being leaked via URLs, for example transaction hashes on transaction detail views. This is not compliant with consent-less tracking.

src/network.ts
```diff
@@ -209,6 +212,19 @@ export async function launchNetwork() {
...
     client.addConsensusChangedListener(async (consensus) => {
         network$.consensus = consensus;
+        if (clientStartTimestamp) {
+            // @ts-expect-error Matomo action queue is not typed
+            window._paq.push([
+                'trackEvent',
+                'Network',
+                `Consensus ${consensus}`,
+                Math.round((Date.now() - clientStartTimestamp) / 100) / 10,
+            ]);
```
danimoh, line 222: Not sure, whether it's possible to process those events per (masked) geo location. Potentially requires using the "Custom Reports" plugin. Otherwise, we might have to add the location to the event on the js side.

danimoh (commented):
Currently, this setup reduces the security of the Wallet by including a script from an external server, and potentially leaks personal data like transaction hashes via page URLs even though we're tracking in consent-less mode.

Also, we would need to check very thoroughly, whether consent-less tracking is realistically possible, as some jurisdictions are very strict. I suggest reading through these pages:
- https://matomo.org/faq/new-to-piwik/how-do-i-use-matomo-analytics-without-consent-or-cookie-banner/
- https://matomo.org/faq/new-to-piwik/how-can-i-still-track-a-visitor-without-cookies-even-if-they-decline-the-cookie-consent/
- https://matomo.org/faq/general/eprivacy-directive-national-implementations-and-website-analytics/

Even though compliant in many parts of the world, my impression is that under the strictest jurisdictions, the consent-less approach taken by this PR is not compliant, even if ignoring the potential leak of personal data via page URLs. For some jurisdictions, we would also need to mention the use of Matomo in the privacy policy and offer an opt-out.

As far as I can see, we are primarily tracking the following information via the Javascript client:
- the initial view
- view changes via the router
- time to first consensus

Instead of tracking the views and view changes, server log tracking might be sufficient, even though less accurate. The consensus event, we might track manually by using the HTTP API directly, which the first page linked above [mentions as being allowed even under strict regulations](https://matomo.org/faq/new-to-piwik/how-do-i-use-matomo-analytics-without-consent-or-cookie-banner/#:~:text=Matomo%20HTTP%20Tracking%20API.), as long as no personal data is being sent. On the other hand, the second link mentions that [under strict regulations, tracking via Javascript is not allowed without consent](https://matomo.org/faq/new-to-piwik/how-can-i-still-track-a-visitor-without-cookies-even-if-they-decline-the-cookie-consent/#:~:text=From%20ePrivacy%20law%20perspective%2C%20JavaScript%20tracking%20is%20treated%20the%20same%20as%20cookie%2Dbased%20tracking%20and%20it%20does%20not%20matter%20if%20the%20data%20processed%20collects%20personal%20data%20or%20not.). Obviously, sending off the event on first consensus would involve Javascript, but maybe the second link was referring to the Javascript client (which also does additional things like fingerprinting) and as the first link mentioned the HTTP API as being allowed, maybe there is some room for interpretation, and tracking that simple consensus event via the HTTP API would be ok without consensus.

But maybe, that first link was in fact only referring to using the HTTP API from a server being ok. In that case, we might have to think about how to track consensus times as a secondary effect, for example by starting to load less important resources only as soon as the consensus is established. That might require the server to measure the time between page load and the additional resources being requested, and then sending of the event from the server.

Another approach which allows measuring the time to consensus would be keeping a websocket connection to nimiq.watch open until consensus is established, to already receive preliminary balance updates and transactions, which would be helpful anyways in countries with slow internet. nimiq.watch would then be able to measure the time such websocket connection was kept open.

Overall, I'd suggest reverting the changes of this PR and only tracking the time to consensus manually via the HTTP API, which also avoids the overhead of loading the Javascript client.

The alternative is of course to actually track only with user consent, which could be opt-in only in an unobtrusive way in the settings, and by a nice invitation to contribute to improving the Wallet after some time of use, for example after sending the third transaction. Even if only a low percentage of user would opt-in, it would give an idea of consensus times around the globe. With consent, we could also get additional insights on usage of the app and also track in the Hub.

src/main.ts
sisou: Done in 1307418dd074933d821250f944504921d1155620

src/matomo.ts
sisou: Done in 1307418dd074933d821250f944504921d1155620.

src/network.ts
sisou: IP anonymization is configured Matomo-wide on our instance.

src/matomo.ts
sisou: The Matomo instance is controlled by us, just like the wallet domain is controlled by us, and the script is loaded via HTTPS. So I'd say that's safe enough.

PR #272, opened 2025-05-26
Nimiq Feedback

onmax:


src/components/modals/NimiqFeedbackModal.vue
```diff
@@ -0,0 +1,107 @@
...
+        onMounted(() => {
+            // Load CSS
+            const cssLink = document.createElement('link');
+            cssLink.rel = 'stylesheet';
+            cssLink.href = 'https://nq-feedback.maximogarciamtnez.workers.dev/widget.css';
+            document.head.appendChild(cssLink);
+
+            // Load JS
+            const script = document.createElement('script');
+            script.src = 'https://nq-feedback.maximogarciamtnez.workers.dev/widget.js';
```
onmax, line 54: Should we download this script on build so we don't have dynamic JS?

PS: probably this url will be changed

PR #290, opened 2026-02-14
fix: point validators API base URLs to workers

onmax:
## Summary
- The wallet was still using deprecated `.nuxt.dev` validator API domains.
- Switched staking API base URLs to Cloudflare Workers endpoints for mainnet and testnet.

## Why
- The legacy domains no longer reliably serve API traffic after migration.

## Notes
- Updated config files:
  - `src/config/config.mainnet.ts`
  - `src/config/config.testnet.ts`
  - `src/config/config.local.ts`


mraveux:
Good to me 👍

PR #291, opened 2026-03-27
Integrate Coinify widget

sisou:
This pull request introduces Coinify as a new buy/sell provider in the app, replacing Moonpay in the buy flow and integrating a new modal for Coinify's widget. The changes update the UI to feature Coinify, adjust logic for provider availability, and add full support for Coinify's widget and event handling.

**Coinify Integration and Buy Flow Updates:**

* Added a new `CoinifyModal.vue` component that embeds the Coinify widget, handles widget events for buy/sell flows, and manages address and currency selection.
* Updated the buy options modal (`BuyOptionsModal.vue`) to replace Moonpay with Coinify.

**Sidebar and Sell Flow Adjustments:**

* Added Coinify as a sell provider in the sidebar, updated the `SellProvider` enum, and ensured Coinify is always available for selling supported currencies.


src/components/modals/CoinifyModal.vue
```diff
@@ -0,0 +1,238 @@
...
+
+            // Pass the relevant addresses for the selected cryptocurrencies, so the user doesn't have to enter
+            // them manually in the widget.
+            widgetAddresses.set('NIM', activeAddress.value!.replace(/\s/g, ''));
+            if (config.enableBitcoin && hasBitcoinAddresses.value && availableExternalAddresses.value.length) {
+                widgetAddresses.set('BTC', availableExternalAddresses.value[0]);
+            }
+            if (config.polygon.enabled && hasPolygonAddresses.value && stablecoin.value) {
+                widgetAddresses.set(`${stablecoin.value.toUpperCase()}POLYGON`, polygonActiveAddress.value!);
+            }
```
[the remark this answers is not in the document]
sisou: The NIM address is sure to be initialized when the CoinifyModal is loaded. The Polygon address is sure to be initialized when `hasPolygonAddresses.value` is true. 

Wont fix.

src/components/modals/CoinifyModal.vue
```diff
@@ -0,0 +1,236 @@
...
+
+        function onWidgetMessage(event: MessageEvent) {
+            if (event.origin !== new URL(config.coinify.widgetUrl).origin) return;
+
+            const msg = event.data;
+
+            // We are only interested in sell confirmation events, where the user must make a transaction to Coinify
+            if (msg.event === 'trade.trade-created' && msg.context.transferIn.medium === 'blockchain') {
+                const payload = msg as {
+                    type: 'event',
```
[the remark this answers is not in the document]
sisou: We are guarding against the `event.origin`, which should be enough. If the Coinify widget ever changes the type of event it fires, this will async throw, which doesn't break the widget itself, it just prevents the automatic send-flow opening.

Wontfix.

src/lib/Countries.ts
```diff
@@ -368,3 +368,143 @@ const SIMPLEX_UNSUPPORTED_COUNTRY_CODES: string[] = [
...
+    'VE',
+    'VI',
+    'VN',
+    'VU',
+    'WS',
+    'XK',
+    'YE',
+    'YT',
+    'ZA',
+    'ZW',
```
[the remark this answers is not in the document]
sisou: The Coinify list was generated with code from their API response, so there is no possibility for human error.

Wontfix.

src/router.ts
```diff
@@ -496,6 +509,14 @@ const routes: RouteConfig[] = [{
             },
             name: RouteName.Simplex,
             meta: { column: Columns.DYNAMIC },
+        }, {
+            path: '/coinify',
+            components: {
+                modal: CoinifyModal,
+            },
+            name: RouteName.Coinify,
+            props: { modal: true },
```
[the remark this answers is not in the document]
sisou: This is wrong. The widget correctly opens with the sell flow selected when coming from the "Sell" button.

For all other widget modal opens, the prop (as defined in the Coinify modal) as a `default` value.

Wontfix.

src/components/modals/CoinifySellInfoModal.vue
```diff
@@ -0,0 +1,181 @@
...
+                            :amount="4.99"
+                            :currency="FiatCurrency.USD"
+                        />
+                    </i18n>
+                </p>
+                <small>{{ $t('+ network fees') }}</small>
+            </main>
+        </PageBody>
+        <PageFooter>
+            <button class="nq-button light-blue" @click="$router.push({ name: 'coinify', params: { flow: 'sell' } })">
```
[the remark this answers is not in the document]
sisou: This is wrong, this works correctly.

Wontfix.

src/components/modals/CoinifyModal.vue
```diff
@@ -0,0 +1,243 @@
...
+                        .replace('usdc-polygon:', 'polygon:');
+                    router.push(`/${paymentUri}`);
+                }
+                if (payload.context.inCurrency === 'USDTPOLYGON') {
+                    // Ensure the user is in USDT mode
+                    useAccountSettingsStore().setStablecoin(CryptoCurrency.USDT);
+                    const paymentUri = payload.context.transferIn.details!.paymentUri
+                        .replace('usdt-polygon:', 'polygon:');
+                    router.push(`/${paymentUri}`);
+                }
```
mraveux, line 162: Nitpicking, I think it makes more sense to make those into `else if`.
```suggestion
                // For NIM and BTC we can use the payment URI as-is
                if (payload.context.inCurrency === 'NIM' || payload.context.inCurrency === 'BTC') {
                    router.push(`/${payload.context.transferIn.details!.paymentUri}`);
                } else if (payload.context.inCurrency === 'USDCPOLYGON') { // For USDC/T, we have to handle some things
                    // Ensure the user is in USDC mode
                    useAccountSettingsStore().setStablecoin(CryptoCurrency.USDC);
                    const paymentUri = payload.context.transferIn.details!.paymentUri
                        .replace('usdc-polygon:', 'polygon:');
                    router.push(`/${paymentUri}`);
                } else if (payload.context.inCurrency === 'USDTPOLYGON') {
                    // Ensure the user is in USDT mode
                    useAccountSettingsStore().setStablecoin(CryptoCurrency.USDT);
                    const paymentUri = payload.context.transferIn.details!.paymentUri
                        .replace('usdt-polygon:', 'polygon:');
                    router.push(`/${paymentUri}`);
                }
```

src/components/modals/CoinifyModal.vue
```diff
@@ -0,0 +1,243 @@
...
+            <div class="flex-spacer"></div>
+        </header>
+        <div class="separator"></div>
+        <iframe
+            :src="widgetUrl"
+            title="Coinify trade widget"
+            allow="camera;fullscreen;accelerometer;gyroscope;magnetometer;payment"
+            allowfullscreen
+            class="widget-iframe flex-grow"
+        />
```
mraveux, line 20: As I was doing some security reviews for the wallet/hub/keyguard I stumbled upon the `sandbox` attribute for iframe. Do you think we should start using it? Something like that maybe?
```suggestion
        <iframe
            :src="widgetUrl"
            title="Coinify trade widget"
            allow="camera;fullscreen;accelerometer;gyroscope;magnetometer;payment"
            allowfullscreen
            class="widget-iframe flex-grow"
            sandbox="allow-scripts allow-same-origin allow-forms allow-popups allow-popups-to-escape-sandbox"
        />
```

mraveux (approved):
The widget works nicely, and I couldn't find much issues. Nice 👍 
Outside of the sandbox attribute, which I don't think is necessary but maybe a "good to have" security-wise, we can imo proceed with the merge and deployment.

src/components/modals/CoinifyModal.vue
sisou: Sure, go ahead. There's no technical difference.

sisou: A `switch` statement would make it even more clear that these are exclusive cases.

src/components/modals/CoinifyModal.vue
sisou: I took the setup code directly from their docs, without much thought: https://coinify-trade-api.readme.io/reference/initial-trade-widget-setup.

If you want to introduce the `sandbox` attribute, you need to make sure the widget still works.

Also it looks to me like the values to add to the `sandbox` attribute defy the reason for sandboxing it :D

mraveux: Yeah I don't think it's needed actually, especially if we set it up to then allow everything to go through x)

PR #311, opened 2026-09-09
Incorporate the watchtower for unstaking and switching validators

mraveux:
### Why

Unstaking and switching validators each need **several transactions signed at once**, with the later ones
only broadcastable after the first has settled on chain. Asking the user to come back hours later to send
the rest is not viable, so the transactions are all signed up front and the remainder handed to the
watchtower, which broadcasts them once the deactivation has matured.

This branch is the wallet half of that. The Hub half (multi-transaction signing plumbing and Ledger
support) is a separate branch, and the Keyguard half is T404-16.

### What

**Watchtower client** — new `src/lib/AlbatrossWatchtower.ts`

- `startUnstaking()` posts `/unstake` with the deactivation hash plus the signed retire and remove-stake
  transactions; `startSwitchValidator()` posts `/switch-validator` with the deactivation hash and the
  signed update-staker transaction.
- Endpoint comes from config per network (`albatrossWatchtower.endpoint`, added to local, testnet and
  mainnet); optional basic auth from `VUE_APP_WT_USERNAME` / `VUE_APP_WT_PASSWORD`.
- Hashes are `0x`-stripped, any non-2xx throws with the server's message, and an unconfigured watchtower
  no-ops with a console warning rather than failing the flow.

**Multi-transaction signing through the Hub** — `src/hub.ts`

- `signUnstakingTransactions()` (layout `unstaking`, exactly 3: deactivate → retire → remove-stake) and
  `signSwitchValidatorTransactions()` (layout `switch-validator`, exactly 2: set-active-stake →
  update-staker). Both **return the signed transactions without broadcasting**, so the caller decides what
  goes on chain now and what goes to the watchtower.
- `sendTransaction()` and the `SIGN_TRANSACTION` listener now accept an array result and send sequentially,
  stopping on the first failed execution.

**Waiting for the chain** — `src/network.ts`, new `src/composables/usePolicy.ts`

- `waitForTransactionConfirmation()` has two modes: inclusion (watch the transaction's state) and
  macro-block confirmation (`requireConfirmed`). The second cannot watch `tx.state` — it is mutated in
  place by `addTransactions` and core does not always refire listeners — so it computes the target macro
  block from the transaction's height and watches the reactive chain height instead.
- `usePolicy()` exposes `BLOCKS_PER_BATCH` / `BATCHES_PER_EPOCH` from `@nimiq/core` (polling briefly for
  the WASM warm-up, no wallet-side fallbacks) and derives `electionBlockAfter` / `macroBlockAfter` from the
  configured genesis, so the wallet's epoch maths lines up with the watchtower's.

**Switch-validator state** — `src/stores/Staking.ts`, new `src/lib/SwitchValidator.ts`

- A switch record (target validator, start height, deactivation hash) is kept in `localStorage` per address,
  exposed through `activeSwitchOperation`, `isSwitchingValidator`, `canManuallyActivateSwitch` and
  `switchTargetLabel`.
- `checkSwitchCompletion()` runs against every stake snapshot and clears the record **only once active
  stake is actually at the target validator** — clearing on any reappearing balance would take away the
  user's manual recovery handle when the stake came back at the old validator.
- `sendImmediateValidatorSwitch()` is that recovery path: a single update-staker with `reactivateAllStake`,
  for when the cooldown has elapsed and the watchtower did not complete the switch.

**UI** (staking pages, validator overlay/modal)

- "Switching to \<validator\>" with the countdown while a switch is in flight, a manual "activate validator"
  button once it can be triggered by hand, `deactivateAll` hidden during a switch, and the unstaking layout.
- Daniel's four commits on top: validator labels, address and images handled consistently across
  `StakingGraphPage` / `StakingInfoPage`, the unstaking validator label passed as `senderLabel`, the
  validator address passed on staking creation, and `@nimiq/hub-api` bumped for multi-tx signing.

### How it was verified

QA ran on `wallet.nimiq-testnet.com` against the ten-flow test plan on T404-15 (regular send, validator
switch, unstaking, instant payout, start staking, add stake, cashlink creation, swap refund, standard
multi-tx signing, checkout — each for both Keyguard and Ledger). Report:
https://app.qase.io/public/report/379bd6f17c667627377c8950bcb27604cd5e0505

### Follow-ups (landed later or still open)

- Follow-up fixes live on `matheo/watchtower-fixes` (T404-173, T404-174, T404-176): gating a validator
  switch while an unstaking is in progress, the inert Switch Validator button, warning the user when the
  watchtower did not accept the job, and verifying pending operations against the watchtower itself.
- `matheo/watchtower-staker-filter` switches `fetchWatchtowerJobsForStaker` to the watchtower's new
  `?staker=` route, once that PR is merged and deployed.
- The switch record still lives in `localStorage` behind a bump counter rather than in reactive store state.
- T404-177: the "Unstake rest" flow does not yet use the watchtower-based multi-tx path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@nimiq/core": "^2.20.0",
     "@nimiq/electrum-client": "https://github.com/nimiq/electrum-client#build",
     "@nimiq/fastspot-api": "^1.10.3",
-    "@nimiq/hub-api": "^1.13.0",
+    "@nimiq/hub-api": "^1.15.0",
     "@nimiq/iqons": "^1.5.2",
     "@nimiq/libswap": "^1.5.2",
     "@nimiq/oasis-api": "^1.1.1",
--- a/src/components/staking/StakingGraphPage.vue
+++ b/src/components/staking/StakingGraphPage.vue
@@ -147,7 +147,6 @@ export default defineComponent({
                         const txs = await sendStaking({
                             transaction: transaction.serialize(),
                             recipientLabel: 'name' in activeValidator.value! ? activeValidator.value.name : 'Validator',
-                            // @ts-expect-error Not typed yet in Hub
                             validatorImageUrl: 'logo' in activeValidator.value!
                                 && !activeValidator.value.hasDefaultLogo
                                 ? activeValidator.value.logo
@@ -190,7 +189,6 @@ export default defineComponent({
                         const txs = await sendStaking({
                             transaction: transaction.serialize(),
                             recipientLabel: 'name' in activeValidator.value! ? activeValidator.value.name : 'Validator',
-                            // @ts-expect-error Not typed yet in Hub
                             validatorAddress: activeValidator.value!.address,
                             validatorImageUrl: ('logo' in activeValidator.value!
                                 && !activeValidator.value.hasDefaultLogo)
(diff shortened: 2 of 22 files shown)


mraveux:
Already merged into master; opened to keep the change set on record.

